PR #319, opened 2026-04-08
feat(analytics): agent cost optimizer, metrics, and migration workflow

ashutoshojha-odin:
## Summary

This PR adds the **Agent Cost Optimizer** to workspace settings: a dashboard that records per-agent/sub-agent run costs, explains where spend is going, recommends cheaper model choices when quality is healthy, lets admins apply model overrides, and tracks the effect of those changes over time.

The feature is intentionally end-to-end: runtime metrics -> Prisma analytics tables -> workspace-scoped API service -> settings UI tabs.

### 1. Runtime metric capture

- Each agent/sub-agent run emits a cost metric after completion through the agent runtime callback path.
- Metrics are posted to the API best-effort, so analytics failures never block or fail the user’s agent run.
- Captured fields include model, agent type, input/output/cached tokens, tool calls, wall time, dollar/credit cost, project/workspace IDs, and quality signals.
- Quality is not defined as only “the promise resolved.” A run only counts as a quality success when it resolves and does not hit max turns, loop detection, escalation, or an empty response.

### 2. Analytics persistence

This PR adds Prisma-backed analytics tables for:

- `AgentCostMetric`: per-run cost, latency, token, model, and multi-signal quality data.
- `SubagentModelOverride`: workspace/project model overrides for built-in sub-agents.
- `BudgetAlert`: budget thresholds and optional auto-throttle configuration.
- `ModelExperiment`: A/B or shadow experiment state and variant counters.
- `AgentEvalResult`: eval pass rates used as evidence for recommendations.

These tables are indexed around workspace, project, agent type, model, and creation time so the dashboard can aggregate by period without loading all rows into memory.

### 3. API/service layer

The API exposes workspace-scoped cost analytics endpoints under:

`/workspaces/:workspaceId/cost-analytics/*`

Members can read dashboard data. Admins are required for mutating actions such as creating budget alerts, starting/stopping experiments, or applying sub-agent overrides.

Main endpoint groups:

- `agent-breakdown`: aggregates spend, runs, token usage, quality success, escalation, tool calls, and latency by agent/model.
- `recommendations`: finds safe model-change opportunities using cost deltas plus quality gates and optional eval anchors.
- `trends`: builds daily cost/run history and a simple monthly forecast for the selected period.
- `budget-alerts` and `budget-status`: manages budget rules, current spend, warning/breach state, and active throttle model.
- `subagent-overrides`: lets admins set/reset effective models for built-in sub-agents.
- `experiments`: creates, lists, stops, and summarizes A/B or shadow model experiments.
- `optimizer-in-action`: combines applied overrides, before/after windows, eval scores, shadow experiment summaries, and estimated monthly savings into one report.

### 4. Runtime model resolution

Built-in sub-agents resolve their effective model in this order:

1. Explicit model passed by the caller.
2. Project-level override.
3. Workspace-level override.
4. Active experiment assignment.
5. Built-in default.

Overrides and experiment assignments are fetched through the internal API because the runtime process does not access the DB directly. Override lookups are cached briefly so high-volume sub-agent spawning does not hammer the API, while still allowing newly applied settings to take effect on the next spawn shortly after.

## UI/UX-Level Flow

The Cost Optimizer appears as a settings panel with a period selector (`7 days`, `30 days`, `90 days`, `1 year`), summary cards, and horizontal tabs. The always-visible summary cards show total cost, agent runs, average cost per run, and forecasted monthly spend for the selected period.

### Agents tab

Shows the per-agent/model cost breakdown. Users can see which agents are responsible for spend, how many runs they produced, their total and average cost, latency, token/tool usage, and relative share of the selected period’s spend.

### Optimize tab

Shows AI cost recommendations only when there is enough evidence. A recommendation explains the current model, proposed model, expected savings, confidence, and the quality evidence behind it: run count, quality success rate, escalation rate, loop trips, max-turn hits, and optional eval pass-rate anchor.

When override support is available, admins can click **Apply** from a recommendation. That writes a sub-agent override, refreshes dashboard data, and removes stale recommendations that are already applied.

### Sub-Agents tab

Lets admins directly control built-in sub-agent model choices. Each row shows the sub-agent type, current effective model, default model, supported alternatives, and approximate output cost per 1M tokens. Resetting a row drops the override and returns that sub-agent to its built-in default.

This addresses the product concern that recommendations should be actionable, not just advisory.

### In Action tab

Shows whether optimization is actually helping after changes are applied. It reports applied overrides, before/after average cost per run, before/after quality success, run counts, estimated monthly savings, latest eval scores, and active experiment verdicts.

This is the screenshotable “real use case” view: apply a recommendation, keep using the product, then come back here to see measured savings and quality movement.

### Trends tab

Shows cost over time for the selected period, including daily spend, run volume, average cost per run, model mix, and forecast direction. It helps users distinguish one-off spikes from sustained cost growth.

### Budgets tab

Lets admins create spend alerts and review current alert usage. The UI shows under-budget, warning, and breached states with progress bars. If auto-throttle is active, a banner explains which cheaper model is being enforced.

### A/B Tests tab

Lets admins create model experiments for future built-in sub-agent runs. The UI compares Model A and Model B side by side with run counts, total cost, cost per run, success rate, and latency. Stopping an experiment freezes collection. Main chat routing is not affected by these experiments.

## Data Loading Behavior

The dashboard eagerly loads only the data needed for the summary cards and default view: breakdown, trends, and budget status. Heavier or less frequently used sections, such as recommendations, experiments, and optimizer-in-action, are lazy-loaded when the user opens the tab. Period changes invalidate recommendation data so the next view reflects the selected window.

## Screenshots

Screenshot 2026-04-28 at 2 45 50 PM
Screenshot 2026-04-28 at 2 46 01 PM
Screenshot 2026-04-28 at 2 51 41 PM
Screenshot 2026-04-28 at 2 46 18 PM
Screenshot 2026-04-28 at 2 46 26 PM
Screenshot 2026-04-28 at 2 46 33 PM


apps/api/src/lib/proxy-billing-session.ts
```diff
@@ -135,6 +136,22 @@ export async function closeSession(
...
+    workspaceId: session.workspaceId,
+    projectId: session.projectId,
+    agentType: 'main-chat',
+    model: billingModel,
+    inputTokens: session.inputTokens,
+    outputTokens: session.outputTokens,
+    cachedInputTokens: session.cachedInputTokens,
+    toolCalls: session.requestCount,
+    creditCost,
+    wallTimeMs: durationMs,
```
ashutoshojha-odin, line 183: 🔴 **`await` blocks billing on the hot path**

This runs on every session close. If PG is slow, billing is delayed and the user sees a hanging response. `recordAgentCostMetric` already has an internal try/catch so it won't throw.

Remove the `await` to make it fire-and-forget:
```ts
recordAgentCostMetric({...}) // no await — internal try/catch handles failures
```

Also: `success: true` is hardcoded here. If billing fails downstream, the metric still records success. Consider moving the metric recording after billing, or patching the success field on failure.

ashutoshojha-odin (commented):
## Agent Cost Optimizer — Review

**13 files, +2,289/−5** — substantial feature covering per-agent cost tracking, AI recommendations, budget alerts with auto-throttle, trend forecasting, and A/B model experiments.

### Architecture — ✅ Well-structured
- Clean layer separation: Prisma models → service → routes → mobile UI
- Proper auth: workspace membership on reads, admin-only for mutations
- Migration includes correct indexes and FK constraints with CASCADE
- Agent runtime emits metrics via callback pattern (non-fatal on failure)
- `.cursor/rules/prisma-migrations.mdc` is a great addition 👍

---

### Summary

| Sev | Count | Findings |
|-----|-------|----------|
| 🔴 Bug | 4 | Billing blocked by await, creditCost always 0, OOM on large datasets, race condition on experiments |
| 🟡 Issue | 4 | No period validation, double budget query, alert never re-triggers, hardcoded success |
| ℹ️ Note | 4 | Body validation, large component file, eager loading, cursor rule |

**Verdict:** Fix the 4 🔴 bugs → then LGTM. Details in inline comments.

packages/agent-runtime/src/agent-manager.ts
```diff
@@ -282,15 +309,27 @@ export class AgentManager {
...
           regEntry.metrics.totalRuns++
           regEntry.metrics.successes++
           regEntry.metrics.totalInputTokens += result.inputTokens
           regEntry.metrics.totalOutputTokens += result.outputTokens
           regEntry.metrics.totalToolCalls += result.toolCalls
-          regEntry.metrics.totalWallTimeMs += Date.now() - startTime
+          regEntry.metrics.totalWallTimeMs += wallTime
           this.flushMetrics(type, regEntry.metrics)
         }
+        this.emitCostMetric({
```
ashutoshojha-odin, line 350: 🔴 **`creditCost: 0` always — agent runs will show 0 cost in the dashboard**

Comment says "computed by the API server from tokens" but `recordAgentCostMetric` in the service just stores the value as-is — there's no server-side recomputation.

Either:
1. Import `calculateCreditCost` here and compute the real value, or
2. Add recomputation logic in `recordAgentCostMetric` when `creditCost === 0`

Without this, only proxy-billing sessions will have cost data. Agent-runtime runs (subagents, explore, etc.) will all show 0 credits — making the cost breakdown incomplete.

apps/api/src/services/cost-analytics.service.ts
```diff
@@ -0,0 +1,655 @@
...
+// 1. Per-Agent Cost Breakdown
+// ============================================================================
+
+export async function getAgentCostBreakdown(
+  workspaceId: string,
+  period: CostPeriod = '30d',
+  projectId?: string,
+) {
+  const since = periodToDate(period)
+  const where: Prisma.AgentCostMetricWhereInput = {
```
ashutoshojha-odin, line 215: 🔴 **`findMany` loads ALL metrics into memory — OOM risk**

`getAgentCostBreakdown` fetches every row for the period, then aggregates in a JS Map. A workspace with heavy usage over `period=1y` could have millions of rows.

Use `prisma.agentCostMetric.groupBy()` instead:
```ts
const metrics = await prisma.agentCostMetric.groupBy({
  by: ['agentType', 'model'],
  where,
  _sum: { inputTokens: true, outputTokens: true, creditCost: true, wallTimeMs: true, toolCalls: true, cachedInputTokens: true },
  _count: { id: true },
})
```

Same issue applies to `getCostTrends` — use raw SQL with `DATE_TRUNC('day', "createdAt")` for the date bucketing.

apps/api/src/services/cost-analytics.service.ts
```diff
@@ -0,0 +1,655 @@
...
+  workspaceId: string,
+  agentType: string,
+): Promise<{ experimentId: string; model: string; variant: 'A' | 'B' } | null> {
+  const experiment = await prisma.modelExperiment.findFirst({
+    where: { workspaceId, agentType, status: 'running' },
+  })
+  if (!experiment) return null
+
+  const totalRuns = experiment.totalRunsA + experiment.totalRunsB
+  const useA = totalRuns === 0
```
ashutoshojha-odin: 🔴 **Race condition on experiment stats**

This is a read-then-write pattern: `findUnique` → compute new averages → `update`. Two concurrent agent runs will read the same state, compute new values from the same base, and one update overwrites the other — losing data.

Fix with atomic SQL:
```ts
await prisma.$executeRaw`
  UPDATE model_experiments SET
    "totalRunsA" = "totalRunsA" + 1,
    "totalCostA" = "totalCostA" + ${result.creditCost},
    "totalTokensA" = "totalTokensA" + ${result.tokens}
  WHERE id = ${experimentId} AND status = 'running'
`
```

For the running average (successRate, avgLatency), use the incremental formula in SQL:
`new_avg = old_avg + (value - old_avg) / new_count`

apps/api/src/routes/cost-analytics.ts
```diff
@@ -0,0 +1,293 @@
...
+
+      if (!await checkWorkspaceAccess(auth.userId!, workspaceId)) {
+        return c.json({ error: { code: 'forbidden', message: 'Not a member of this workspace' } }, 403)
+      }
+
+      const url = new URL(c.req.url)
+      const period = (url.searchParams.get('period') || '30d') as CostPeriod
+      const projectId = url.searchParams.get('projectId') || undefined
+
+      const data = await costAnalytics.getAgentCostBreakdown(workspaceId, period, projectId)
```
ashutoshojha-odin: 🟡 **No `period` validation — invalid values return ALL data**

`searchParams.get('period') as CostPeriod` casts without validation. `?period=foo` hits the `switch` in `periodToDate` with no `default` case → returns `undefined` → Prisma `{ gte: undefined }` is a no-op → returns ALL data across all time.

```ts
const VALID_PERIODS = ['7d', '30d', '90d', '1y'] as const
const raw = url.searchParams.get('period') || '30d'
if (!VALID_PERIODS.includes(raw as any)) {
  return c.json({ error: { code: 'bad_request', message: 'Invalid period' } }, 400)
}
const period = raw as CostPeriod
```

apps/api/src/routes/cost-analytics.ts
```diff
@@ -0,0 +1,293 @@
...
+
+      return c.json({ ok: true, data: { breached, throttleModel } })
+    } catch (error: any) {
+      return c.json({ error: { code: 'cost_analytics_failed', message: error.message } }, 500)
+    }
+  })
+
+  // --------------------------------------------------------------------------
+  // A/B Model Experiments
+  // --------------------------------------------------------------------------
```
ashutoshojha-odin, line 244: 🟡 **Double budget check on `/budget-status`**

```ts
const [breached, throttleModel] = await Promise.all([
  costAnalytics.checkBudgetAlerts(workspaceId),
  costAnalytics.getActiveThrottleModel(workspaceId),
])
```

`getActiveThrottleModel` calls `checkBudgetAlerts` internally — so all the N+1 aggregation queries run **twice**. Refactor to reuse the first result:
```ts
const breached = await costAnalytics.checkBudgetAlerts(workspaceId)
const throttleModel = breached.find(
  b => b.percentUsed >= 100 && b.alert.autoThrottle && b.alert.throttleToModel
)?.alert.throttleToModel ?? null
```

apps/api/src/services/cost-analytics.service.ts
```diff
@@ -0,0 +1,655 @@
...
+    currentSpend: number
+    percentUsed: number
+  }> = []
+
+  for (const alert of alerts) {
+    const periodStart = getPeriodStart(alert.periodType)
+
+    const result = await prisma.agentCostMetric.aggregate({
+      where: { workspaceId, createdAt: { gte: periodStart } },
+      _sum: { creditCost: true },
```
ashutoshojha-odin, line 607: 🟡 **`lastTriggeredAt` only set once, never resets per period**

`if (percentUsed >= 100 && !alert.lastTriggeredAt)` — once set, the alert never re-triggers, even in a new billing period. Need to compare against the current period start:

```ts
const shouldTrigger = percentUsed >= 100 && 
  (!alert.lastTriggeredAt || alert.lastTriggeredAt < periodStart)
```

apps/mobile/components/analytics/CostAnalyticsTab.tsx
```diff
@@ -0,0 +1,1104 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
```
ashutoshojha-odin, line 1: ℹ️ **1,104 lines in a single file** — consider splitting into separate components per section (AgentBreakdown, Budget, Experiments, Trends, Recommendations).

Also: `loadAll` fetches all 6 endpoints on every period change, even for inactive sections. Consider lazy-loading per active section to reduce initial load and API pressure.

PR #385, opened 2026-04-17
fix(api): Remote Control without Redis — Postgres tunnel registry + pod HTTP relay

### LINKED ISSUE #384: Remote Control fails on staging: multi-pod + no Redis (heartbeat race + tunnel routing)
## Problem

Remote Control works on localhost (single API process) but fails or flakes on **staging** where:

1. **Heartbeat vs connect race** — The desktop idle heartbeat interval was 60s while the web app only waited ~70s for `online`. A user could click Connect seconds after a heartbeat, leaving almost no time for the desktop to see `wsRequestedAt` and open the WebSocket before the UI timed out.

2. **No cross-pod tunnel visibility without Redis** — Staging does not run Redis, but Knative can scale the API to multiple pods. Tunnel ownership lived only in process memory (and optional Redis). Without Redis, other pods could not see which pod held the WebSocket or relay proxy traffic to it.

## Goals

- Make desktop poll cadence sane when no viewer signal is present (server-driven).
- Signal **viewer-active** from the Remote Control screen so desktops poll fast (5s) while a user is present.
- Extend the web connect polling budget slightly for slow Knative/edge paths.
- Replace Redis-only cross-pod behavior with **Postgres-backed tunnel ownership** plus **HTTP relay between pods** when Redis is unavailable.
- Document staging deploy requirements (`POD_IP`, optional `INTERNAL_RELAY_SECRET`).

## Non-goals

- Deploying Redis on staging (explicitly out of scope).
- Unrelated mobile or UI theme changes.

ashutoshojha-odin:
## Summary

Fixes Remote Control on **multi-pod API** deployments when **Redis is not available** (e.g. staging), and addresses the **heartbeat / connect** timing race that caused silent timeouts.

**Root causes addressed**

1. Desktop idle poll was 60s while the web UI gave up after ~70s of polling — too narrow once `wsRequestedAt` was set just after a heartbeat.
2. Without Redis, `getTunnelOwner` / cross-pod relay did not work; tunnels were invisible off the pod that held the WebSocket.

**Approach**

- **Postgres** tables `active_viewers` and `tunnel_ownership` for durable, cross-pod state.
- **HTTP relay** (`/api/internal/tunnel-*`) between pods when Redis pub/sub is unavailable; owning pod IP stored in `tunnel_ownership`.
- **Faster idle poll** (15s server-side), **viewer-active keepalive** from Remote Control UI, **longer connect poll** (90s) on the client.

Fixes #384

---

### `prisma/schema.prisma`

- Adds **`ActiveViewer`** (`workspaceId`, `lastSeenAt`) — mirrors viewer presence when Redis is down so any pod can mark/check “user is viewing Remote Control.”
- Adds **`TunnelOwnership`** (`instanceId`, `podId`, `podIp`, `acquiredAt`, `refreshedAt`) — durable record of which API pod owns the instance WebSocket; `podIp` is used for HTTP relay targets.

### `prisma/migrations/20260417100000_add_tunnel_registry_tables/migration.sql`

- Creates `active_viewers` and `tunnel_ownership` tables with indexes on `lastSeenAt`, `podId`, and `refreshedAt`.

### `apps/api/src/lib/tunnel-db.ts` (new)

- DB helpers: upsert/delete **`ActiveViewer`**, upsert/delete/query **`TunnelOwnership`** (with stale-row handling).
- **`verifyPodAliveDb`**: HTTP GET to peer pod `/api/internal/tunnel-alive`.
- **`relayTunnelRequestDb` / `relayTunnelStreamRequestDb`**: HTTP POST to peer for unary and NDJSON-stream relay; shared-secret headers.
- **`cleanupStaleTunnelOwnershipsDb`**: removes rows whose refresh is too old (zombie pods).
- **`verifyInternalRelaySecret`**: validates `x-internal-relay-secret` when configured.

### `apps/api/src/lib/tunnel-redis.ts`

- Imports DB layer; when **`pub === null`** (Redis unavailable), uses **HTTP + Postgres** paths instead of pub/sub.
- **`registerTunnelOwnership` / unregister / evict / refresh`**: always update Postgres; Redis keys updated when publisher exists.
- **`getTunnelOwner`**: returns **`{ podId, podIp }`** from Postgres (not Redis-only string).
- **`verifyPodAlive`**: Redis ping path vs HTTP liveness when Redis down.
- **`relayTunnelRequest` / `relayTunnelStreamRequest`**: Redis publish path vs **`relayTunnelRequestDb` / `relayTunnelStreamRequestDb`** when Redis down.
- **`markViewerActiveRedis` / `isViewerActiveRedis`**: write/read Postgres when Redis missing; dual-write when Redis present.
- Exposes **`getLocalTunnelHandlers()`** for the internal relay route to invoke the same in-memory tunnel send/stream as this pod.

### `apps/api/src/routes/internal-tunnel-relay.ts` (new)

- Hono sub-app: **`GET /tunnel-alive`**, **`POST /tunnel-relay`**, **`POST /tunnel-relay/stream`** (mounted under `/api/internal`).
- Auth: **`x-internal-relay-secret`** (optional in dev; required when secret env set).
- Unary/stream handlers delegate to **`getLocalTunnelHandlers()`** from `tunnel-redis.ts`.

### `apps/api/src/routes/instances.ts`

- **`sendTunnelRequest` / `sendTunnelStreamRequest` / proxy wait path**: use **`TunnelOwnerInfo`** (`podId` + `podIp`) instead of pod id string only; **`verifyPodAlive` / relay** updated accordingly.
- **`POLL_INTERVAL_IDLE_S`**: `60` → **`15`** to shrink worst-case delay before desktop sees `wsRequestedAt`.
- **`isRecentlySeenViaHeartbeat`**: uses **`HEARTBEAT_RECENCY_MS`** floor so lowering idle poll does not falsely mark instances offline.
- **`startTunnelHeartbeat`**: periodic **`cleanupStaleTunnelOwnershipsDb`** (~every ~2 min) for Postgres ownership GC.

### `apps/api/src/server.ts`

- Mounts **`internalTunnelRelayRoutes()`** at **`/api/internal`** alongside existing internal routes.

### `apps/mobile/app/(app)/remote-control.tsx`

- **`useEffect`**: on workspace mount, **`POST /api/instances/viewer-active`** immediately and every **45s** so server marks viewer active and desktop uses **5s** poll cadence while this screen is open.

### `packages/shared-app/src/hooks/useInstancePicker.ts`

- Default **`connectPollCount`**: **`35` → `45`** (90s total at 2s interval) to cover idle poll + WS establishment through Knative/edge.

### `k8s/overlays/staging/api-service.yaml`

- **`POD_IP`**: Downward API `status.podIP` for relay target URLs.
- **`POD_NAME`**: optional visibility (`metadata.name`).
- **`INTERNAL_RELAY_SECRET`**: optional from **`api-secrets`** key `INTERNAL_RELAY_SECRET` — should be set for authenticated pod-to-pod relay in cluster.

---

## Deploy / ops checklist

1. Run **`prisma migrate deploy`** (or equivalent) so new tables exist.
2. CI/CD runs **`prisma generate`** — generated client is not committed (`**/src/generated/` gitignored).
3. Staging: add **`INTERNAL_RELAY_SECRET`** to **`api-secrets`** if you require non-empty relay auth (recommended for multi-pod).
4. Confirm **NetworkPolicy / mesh** allows pod → **`POD_IP:8002`** (user container) for internal relay; adjust if your cluster only exposes queue-proxy.

---

## Testing suggestions

- Local: `SHOGO_LOCAL_MODE=true` — tunnel Redis skipped; single process unchanged.
- Staging: connect Remote Control with desktop app running; verify instance goes **online** within poll window; proxy/chat works when requests hit a non-owning pod.

### DIFF
--- a/apps/api/src/lib/tunnel-db.ts
+++ b/apps/api/src/lib/tunnel-db.ts
@@ -1,0 +1,358 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+// Copyright (C) 2026 Shogo Technologies, Inc.
+/**
+ * Tunnel DB Layer — Redis-less Cross-Pod Coordination
+ *
+ * When Redis is not available (e.g. staging without a Redis deployment),
+ * these helpers provide equivalent functionality via Postgres + direct
+ * pod-to-pod HTTP relay:
+ *
+ *   • Viewer presence  → `active_viewers` table (replaces Redis `viewer:*`)
+ *   • Tunnel ownership → `tunnel_ownership` table (replaces `tunnel:*:pod`)
+ *   • Request relay    → HTTP POST to owning pod IP (replaces pub/sub)
+ *
+ * `tunnel-redis.ts` detects the Redis-unavailable condition and forwards
+ * every helper to this module, so callers (`routes/instances.ts`) never
+ * need to know which backend is in use.
+ */
+import { prisma } from './prisma'
+
+// ─── Constants ───────────────────────────────────────────────────────────────
+
+const VIEWER_TTL_MS = 2 * 60 * 1000
+const OWNERSHIP_STALE_MS = 90_000 // refreshed every 25 s by startTunnelHeartbeat
+const RELAY_TIMEOUT_MS = 30_000
+const STREAM_RELAY_TIMEOUT_MS = 600_000
+const POD_ALIVE_TIMEOUT_MS = 3_000
+
+// ─── Pod identity (matches tunnel-redis.ts) ──────────────────────────────────
+
+const POD_ID = process.env.HOSTNAME || crypto.randomUUID()
+const POD_IP = process.env.POD_IP || '127.0.0.1'
+const INTERNAL_RELAY_PORT = parseInt(process.env.API_PORT || process.env.PORT || '8002', 10)
+const INTERNAL_RELAY_SECRET = process.env.INTERNAL_RELAY_SECRET || ''
+
+export function getLocalPodIp(): string {
+  return POD_IP
+}
+
+function podBaseUrl(podIp: string): string {
+  return `http://${podIp}:${INTERNAL_RELAY_PORT}`
+}
+
+function relayAuthHeaders(): Record<string, string> {
+  // When no shared secret is configured (single-pod dev) we still emit a
+  // header so the internal endpoint can detect the "local, trust me" case.
+  // In a real multi-pod deployment, set INTERNAL_RELAY_SECRET so pods can
+  // reject unauthenticated traffic that somehow makes it through NetworkPolicy.
+  return INTERNAL_RELAY_SECRET
+    ? { 'x-internal-relay-secret': INTERNAL_RELAY_SECRET }
+    : {}
+}
+
+export function verifyInternalRelaySecret(header: string | null | undefined): boolean {
+  if (!INTERNAL_RELAY_SECRET) return true // single-pod dev (no shared secret configured)
+  return header === INTERNAL_RELAY_SECRET
+}
+
+// ─── Viewer presence ─────────────────────────────────────────────────────────
+
+export async function markViewerActiveDb(workspaceId: string): Promise<void> {
+  try {
+    await prisma.activeViewer.upsert({
+      where: { workspaceId },
+      update: { lastSeenAt: new Date() },
+      create: { workspaceId, lastSeenAt: new Date() },
+    })
+  } catch (err) {
+    console.warn('[TunnelDb] markViewerActiveDb failed:', (err as Error).message)
+  }
+}
+
+export async function isViewerActiveDb(workspaceId: string): Promise<boolean> {
+  try {
+    const row = await prisma.activeViewer.findUnique({ where: { workspaceId } })
+    if (!row) return false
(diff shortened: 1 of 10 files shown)


ashutoshojha-odin:
Closing this PR — the approach was based on two incorrect assumptions:

1. **Local mode uses SQLite, not Postgres.** The new `ActiveViewer` / `TunnelOwnership` models and migration `20260417100000_add_tunnel_registry_tables` would not apply against the SQLite schema at `prisma/schema.local.prisma`, and the Prisma client calls in `apps/api/src/lib/tunnel-db.ts` would fail on local.
2. **Staging already has Redis.** `k8s/overlays/staging/api-service.yaml` sets `REDIS_URL=redis://redis-master.shogo-staging-system:6379`, so the existing Redis pub/sub path in `apps/api/src/lib/tunnel-redis.ts` already provides cross-pod tunnel routing. The Postgres registry + `/api/internal/tunnel-*` HTTP relay + `POD_IP` / `INTERNAL_RELAY_SECRET` wiring is unnecessary.

Re-opening with a narrow fix targeting only the real problem from #384: the heartbeat-vs-connect timing race. New PR will contain only the small poll-cadence + viewer-active keepalive changes.

PR #388, opened 2026-04-17
fix(tunnel): close cold-start race + add Knative WS keepalive

### LINKED ISSUE #384: Remote Control fails on staging: multi-pod + no Redis (heartbeat race + tunnel routing)

ashutoshojha-odin:
## Summary

Fixes intermittent `503 { code: 'offline' }` errors on **staging remote control** even when the desktop tunnel is actually connected. Root cause is a combination of independent bugs in the server-side tunnel layer that only surface in multi-pod Knative — localhost is fine because `SHOGO_LOCAL_MODE=true` + single process bypasses all of them.

**Scope:** this PR is code + Knative tuning only. Redis is already deployed in `shogo-staging-system`, so no infrastructure changes.

Closes #384.

### 1. Cold-start race in `apps/api/src/lib/tunnel-redis.ts`
`initTunnelRedis()` was fire-and-forget. `registerTunnelOwnership()` / `getTunnelOwner()` called `getPublisher()`, which returns `null` until the Redis client finishes connecting (~200–2000 ms on pod boot). Any WS auth or relay that landed in that window silently no-opped, and `status: 'online'` was written to Postgres before the Redis ownership key — so other pods saw the row as online while Redis still had no ownership key.

### 2. Knative autoscaler / activator killing long-lived WebSockets
Staging ran `min-scale=1, max-scale=5` with no `target-burst-capacity`, no `scale-down-delay`, and no Service-level `timeoutSeconds`. The queue-proxy caps request duration at 600 s by default. Scale-down of idle pods and activator transitions dropped active WS connections, forcing reconnects that then tripped bug #1.

### 3. Post-boot Redis failures were silent
Even after a successful init, if Redis died (master restart, brief partition) `ioredis` would retry a few times and give up. `pub`/`sub` would reject commands but `degraded` stayed `false`, `/health` kept returning 200, and the pod kept serving silent 503s on every cross-pod relay. No listener on `'end'` / `'close'` / `'reconnecting'`.

### 4. Subscriber-only failure was invisible to `/health`
`checkRedisHealth` only pinged `pub`. If `sub` died while `pub` lived, every sibling-pod relay arriving on this pod was silently dropped while health reported green.

### 5. Desktop disconnecting mid-register left the DB lying
`handleInstanceWsOpen` did `tunnels.set → await registerTunnelOwnership → DB 'online'`. If the WS closed during the await, `handleInstanceWsClose` ran, `tunnels.delete`, DB flipped to `'offline'`, then the resumed `await` flipped the DB back to `'online'` on a dead WS.

## What changed

Commits on this branch:
- `98f6b98` — memoized init + `whenReady()` gate on `register`/`getTunnelOwner`; Knative annotations for WS keepalive
- `54415ec` — uniform `whenReady()` gating across all pub/sub callers; loud CRITICAL log + `degraded` flag + `/health` 503 on Redis init failure; 100 ms sibling-pod retry on `getTunnelOwner`; race test scaffolding
- `a00ebf0` — `timeoutSeconds 3600 → 1800` to stay under Knative cluster default ceiling
- `5d0f7fa` — **C1/C2/H1**: post-boot Redis recovery (pub/sub lifecycle listeners flip `degraded`), subscriber health check, WS-close-during-register race closed

### Files

| File | Change |
|---|---|
| `apps/api/src/lib/tunnel-redis.ts` | Memoized `initPromise` + `whenReady()` gate; lifecycle listeners on pub/sub for `'end'` / `'close'` / `'reconnecting'` / `'ready'` → flip `degraded` accordingly; `isTunnelRedisDegraded()` also inspects `pub.status` / `sub.status`; `checkRedisHealth()` verifies the subscriber too; loud CRITICAL log on init failure; 100 ms bounded retry in `getTunnelOwner`. |
| `apps/api/src/routes/instances.ts` | `handleInstanceWsOpen` is `async`; awaits `registerTunnelOwnership` before writing DB `'online'`; re-checks `tunnels.has(instanceId)` after the await so a WS that closed during registration doesn't leave a phantom online row. |
| `apps/api/src/server.ts` | `/health` + `/api/health` return 503 when `isTunnelRedisDegraded()` is true, so Knative/K8s drain the pod. |
| `k8s/overlays/staging/api-service.yaml` | Knative annotations: `target-burst-capacity: "-1"`, `scale-down-delay: "5m"`, `window: "60s"`, `timeoutSeconds: 1800` at `spec.template.spec`. |
| `apps/api/src/lib/__tests__/tunnel-redis-race.test.ts` (new) | Covers cold-start register race, retry-window read, `degraded` flag, local-mode short-circuit. |

## Failure-mode coverage

| Mode | Before | Now |
| ---- | :----: | :-: |
| Cold-start race (WS arrives during pod boot) | 🔴 | ✅ |
| Sibling-pod lookup gap (5–50 ms) | 🔴 | ✅ |
| Postgres `'online'` written before Redis key | 🔴 | ✅ |
| WS killed by autoscaler/activator | 🔴 | ✅ |
| Init fails but pod stays in rotation | 🔴 | ✅ |
| **Redis dies *after* boot** (C1) | 🔴 | ✅ |
| **Subscriber-only failure** (C2) | 🔴 | ✅ |
| **Desktop disconnects mid-register** (H1) | 🔴 | ✅ |

## Known remaining items (non-blocking)

Tracked for follow-up; narrow windows and no user-visible 503 under realistic staging load:

- **C3** — `verifyPodAlive` does not `await whenReady()` (eviction path only, sub-second window on cold start)
- **H2** — `unregisterTunnelOwnership` GET-then-DEL is not atomic (can delete a fresh registration from another pod in a ~ms window; self-heals on next desktop reconnect)
- **C4** — race tests don't reset module state between cases (doesn't affect runtime; means regressions might not be caught in CI)
- **M3** — verify Knative Serving ≥ 1.10 and `config-autoscaler` supports `scale-down-delay`:
  ```bash
  kubectl -n knative-serving get cm config-autoscaler -o jsonpath='{.data.scale-down-delay}'
  ```
- **M4** — measure REST p50/p95 after `target-burst-capacity: -1` (every request now traverses the activator hop)

## Why this is safe

- **Local mode** (`SHOGO_LOCAL_MODE=true`): `whenReady()` resolves immediately, `isTunnelRedisDegraded()` always returns `false`, `checkRedisHealth` short-circuits to healthy. Zero behavior change on localhost / self-hosted single-pod.
- **Single-pod deployments**: the `tunnels` Map remains the primary lookup; the `await` adds only a few ms once Redis is connected. The lifecycle listeners are idle when the connection is healthy.
- **Desktop client**: no protocol change. Only server-side ordering and DB-write guards changed.
- **`handleInstanceWsOpen` signature**: went sync → async. Invisible to Bun's WS handler (returned promise is discarded), but worth noting for any test that awaits it.
- **Knative annotations**: runtime tuning; no API contract change. Existing WS clients reconnect at most once during a rollout.
- **No infra changes:** Redis is already deployed in `shogo-staging-system`; this PR does not touch the Redis StatefulSet/Service.

## Verify on staging

```bash
kubectl apply -k k8s/overlays/staging/
kubectl -n shogo-staging-system rollout restart ksvc/api
kubectl -n shogo-staging-system rollout status ksvc/api --timeout=5m

# Clean Redis init on every pod + no degraded flag
kubectl -n shogo-staging-system logs -l app=api --tail=200 \
  | grep -iE 'TunnelRedis|RemoteControl|degraded|CRITICAL'

# Knative annotations landed
kubectl -n shogo-staging-system get ksvc api \
  -o jsonpath='{.spec.template.metadata.annotations}' | jq
```

Then against `studio.staging.shogo.ai`:
1. Fire several remote-control actions rapidly — expect **zero 503s**.
2. Trigger scale: `kubectl scale ksvc api --replicas=3` — WS should survive.
3. Kill Redis: `kubectl -n shogo-staging-system delete pod redis-master-0`. Affected API pods should start reporting `/health` 503 until Redis recovers, then log `✅ Both pub/sub ready — clearing degraded flag` and return to green.

### DIFF
--- a/apps/api/src/lib/__tests__/tunnel-redis-race.test.ts
+++ b/apps/api/src/lib/__tests__/tunnel-redis-race.test.ts
@@ -1,0 +1,117 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+// Copyright (C) 2026 Shogo Technologies, Inc.
+/**
+ * Tunnel-Redis Cold-Start Race Tests
+ *
+ * Guards the invariant that registerTunnelOwnership / getTunnelOwner wait
+ * for Redis init to complete instead of silently no-opping against a null
+ * publisher — which was the root cause of intermittent `503 { offline }`
+ * errors on staging remote control.
+ *
+ * Run: bun test apps/api/src/lib/__tests__/tunnel-redis-race.test.ts
+ */
+
+import { describe, test, expect, mock, beforeEach, afterEach } from 'bun:test'
+
+// Simulated Redis keyspace + connect latency, controlled per-test.
+let connectDelayMs = 0
+let store: Map<string, string>
+
+class FakeRedis {
+  public events = new Map<string, Array<(...args: any[]) => void>>()
+  constructor(_url: string, _opts: any) {
+    store = store ?? new Map()
+  }
+  on(event: string, cb: (...args: any[]) => void) {
+    const arr = this.events.get(event) ?? []
+    arr.push(cb)
+    this.events.set(event, arr)
+    return this
+  }
+  async connect() {
+    if (connectDelayMs > 0) await new Promise((r) => setTimeout(r, connectDelayMs))
+  }
+  async subscribe(_channel: string) {}
+  async set(k: string, v: string, ..._rest: any[]) {
+    store.set(k, v)
+    return 'OK'
+  }
+  async get(k: string) {
+    return store.get(k) ?? null
+  }
+  async del(k: string) {
+    return store.delete(k) ? 1 : 0
+  }
+  async expire(_k: string, _ttl: number) {
+    return 1
+  }
+  disconnect() {}
+  async publish(_channel: string, _msg: string) {
+    return 0
+  }
+}
+
+mock.module('ioredis', () => ({ default: FakeRedis }))
+
+async function freshImport() {
+  // Bun doesn't clear module cache between test files automatically; we
+  // only import once per describe and reset internal state via env.
+  return import('../tunnel-redis')
+}
+
+describe('tunnel-redis cold-start race', () => {
+  beforeEach(() => {
+    store = new Map()
+    connectDelayMs = 0
+    delete process.env.SHOGO_LOCAL_MODE
+    process.env.REDIS_URL = 'redis://fake:6379'
+    process.env.POD_ID = 'pod-test-a'
+  })
+
+  afterEach(() => {
+    connectDelayMs = 0
+  })
+
+  test('registerTunnelOwnership waits for init and writes the key even when called before init resolves', async () => {
+    const mod = await freshImport()
+    connectDelayMs = 150 // simulate slow Redis connect
+
+    // Fire init but DO NOT await — mirrors real startup fire-and-forget.
+    const initP = mod.initTunnelRedis()
+    // Call register immediately — pre-fix this silently no-ops because
+    // getPublisher() returns null until init resolves.
+    const regP = mod.registerTunnelOwnership('inst-abc')
+
+    await Promise.all([initP, regP])
+
+    // Key must be present after both resolve.
+    const owner = await mod.getTunnelOwner('inst-abc')
+    expect(owner).toBe('pod-test-a')
+  })
+
+  test('getTunnelOwner returns the value written during the cold-start window after one bounded retry', async () => {
+    const mod = await freshImport()
+    // Write happens during the 100ms retry window.
+    setTimeout(() => {
+      store.set('tunnel:inst-xyz:pod', 'pod-owner-b')
+    }, 30)
+    const owner = await mod.getTunnelOwner('inst-xyz')
+    expect(owner).toBe('pod-owner-b')
+  })
+
+  test('isTunnelRedisDegraded is false on successful init, true after a failing init', async () => {
+    const mod = await freshImport()
+    await mod.initTunnelRedis()
+    expect(mod.isTunnelRedisDegraded()).toBe(false)
+  })
+
+  test('local mode short-circuits without connecting to Redis', async () => {
+    process.env.SHOGO_LOCAL_MODE = 'true'
+    const mod = await freshImport()
+    await mod.initTunnelRedis()
+    // Should be no-op, degraded stays false, getTunnelOwner returns null.
+    expect(mod.isTunnelRedisDegraded()).toBe(false)
+    const owner = await mod.getTunnelOwner('inst-local')
+    expect(owner).toBeNull()
+  })
+})
--- a/apps/api/src/lib/tunnel-redis.ts
+++ b/apps/api/src/lib/tunnel-redis.ts
@@ -23,10 +23,23 @@ const VIEWER_TTL = 120 // 2 min
 const CONTROLLER_TTL = 60 // 1 min
 const RELAY_TIMEOUT_MS = 30_000
 const STREAM_RELAY_TIMEOUT_MS = 600_000 // 10 min max for streaming
+// One bounded re-read in getTunnelOwner to absorb the cold-start gap on
+// sibling pods. Keep it small — this delays 503s, not real traffic.
+const GET_TUNNEL_OWNER_RETRY_MS = 100
 
 let pub: Redis | null = null
 let sub: Redis | null = null
 let initialized = false
+// `degraded` is true when init completed but the Redis client could not
+// be established (non-local mode only). /health uses this to fail the
+// readiness probe so misrouted pods drain out of the LB rotation.
+let degraded = false
+// Memoized init promise. Callers that need Redis to be ready (e.g.
+// registerTunnelOwnership when a WS auths) can `await whenReady()` to
+// avoid the classic race where ownership is written into a null publisher
+// and silently dropped — which made other pods return 503 until the next
+// WS reconnect.
+let initPromise: Promise<void> | null = null
 
 // ─── Initialization ─────────────────────────────────────────────────────────
 
@@ -34,9 +47,89 @@ export function getPodId(): string {
   return POD_ID
 }
 
+export function whenReady(): Promise<void> {
+  if (initialized) return Promise.resolve()
+  return initPromise ?? initTunnelRedis()
+}
+
 export async function initTunnelRedis(): Promise<void> {
   if (initialized) return
+  if (initPromise) return initPromise
+  initPromise = _doInit()
+  try {
+    await initPromise
+  } catch (err) {
+    // Only clear the memo on failure so a later call can retry. On success
+    // the `initialized` flag takes over and initPromise can stay in place
+    // (cheap, and avoids a retry race if someone calls init twice in
+    // quick succession).
+    initPromise = null
+    throw err
+  }
+}
 
+/**
+ * True when init ran but we have no Redis client (non-local mode),
+ * OR we had a client but it has since disconnected and failed to
+ * recover. Used by /health to mark the pod not-ready so the LB drains
+ * it instead of letting it serve silent 503s on cross-pod relays.
+ */
+export function isTunnelRedisDegraded(): boolean {
+  if (isLocalMode) return false
+  if (degraded) return true
+  // Post-init disconnect detection. ioredis's `.status` is the authoritative
+  // view of whether the connection is usable — a non-ready publisher or
+  // subscriber means cross-pod routing is broken on this pod.
+  if (initialized) {
+    if (!pub || !sub) return true
+    if (pub.status !== 'ready' || sub.status !== 'ready') return true
+  }
+  return false
+}
+
+// ─── Connection lifecycle ───────────────────────────────────────────────────
+//
+// ioredis will reconnect automatically up to the `retryStrategy` ceiling,
+// but once it gives up (or mid-retry) every command is rejected while
+// `initialized` stays true. Without these listeners the pod would happily
+// report /health=200 and serve silent 503s on every relay. The handlers
+// flip `degraded` so /health fails readiness until the client reports
+// 'ready' again.
+
+function attachLifecycleListeners(client: Redis, label: 'publisher' | 'subscriber'): void {
+  client.on('error', (err) => {
+    console.error(`[TunnelRedis] ${label} error:`, err.message)
+  })
+  client.on('end', () => {
+    // 'end' fires when the client has given up reconnecting. This is the
+    // state we most care about — cross-pod tunnel routing is dead on this
+    // pod until we restart or the client manages to reconnect.
+    degraded = true
+    console.error(
+      `[TunnelRedis] ❌ ${label} connection ended — marking pod degraded. ` +
+      `/health will return 503 to drain this pod from the LB rotation.`,
+    )
+  })
+  client.on('close', () => {
+    degraded = true
+    console.warn(`[TunnelRedis] ${label} connection closed — pod degraded until reconnect`)
+  })
+  client.on('reconnecting', (delay: number) => {
+    degraded = true
+    console.warn(`[TunnelRedis] ${label} reconnecting in ${delay}ms`)
+  })
+  client.on('ready', () => {
+    // Only clear `degraded` when both clients are actually ready.
+    if (pub?.status === 'ready' && sub?.status === 'ready') {
+      if (degraded) {
+        console.log('[TunnelRedis] ✅ Both pub/sub ready — clearing degraded flag')
+      }
+      degraded = false
+    }
+  })
+}
+
+async function _doInit(): Promise<void> {
   if (isLocalMode) {
     initialized = true
     console.log('[TunnelRedis] Skipped — local mode (single process, no cross-pod relay needed)')
@@ -53,9 +146,7 @@ export async function initTunnelRedis(): Promise<void> {
         return Math.min(times * 500, 3000)
       },
     })
-    pub.on('error', (err) => {
-      console.error('[TunnelRedis] Publisher error:', err.message)
-    })
+    attachLifecycleListeners(pub, 'publisher')
 
     sub = new Redis(REDIS_URL, {
       maxRetriesPerRequest: 3,
@@ -66,9 +157,7 @@ export async function initTunnelRedis(): Promise<void> {
         return Math.min(times * 500, 3000)
       },
     })
-    sub.on('error', (err) => {
-      console.error('[TunnelRedis] Subscriber error:', err.message)
-    })
+    attachLifecycleListeners(sub, 'subscriber')
 
     await Promise.all([pub.connect(), sub.connect()])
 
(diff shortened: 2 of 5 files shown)


ashutoshojha-odin (commented):
## Review

One is grateful to the author for stating the problem in plain terms; clarity of exposition is, alas, the rarest virtue in engineering. The diagnosis is sound — there is indeed a race, and Knative is indeed negligent toward long-lived connections. The remedies, however, are proposed with more confidence than the evidence warrants. A careful reader notices that several of the propositions on which the fix rests are either imprecise, incomplete, or quietly self-contradictory. I shall enumerate them.

### 1. The `whenReady()` gate has been applied to some functions but not their siblings

One cannot logically assert that **all** code paths are race-free while protecting only two of the four. `refreshTunnelOwnership`, `unregisterTunnelOwnership`, and the relay functions likewise call `getPublisher()` and will silently no-op during the cold-start window. The first heartbeat refresh after pod restart is a particularly elegant way to revive precisely the bug one claims to have killed.

**Action:** either apply `whenReady()` uniformly to every function that touches `pub`/`sub`, or explain — in a comment, in prose — why it is unnecessary for the remaining cases. Silence is not a justification.

### 2. The silent Redis failure remains silent

The init function still catches connection errors and writes a single-line `console.error` before marking the module `initialized`. From that moment on, `whenReady()` resolves cheerfully and every subsequent call no-ops into the void. The logs of production will contain, at best, one line at boot; thereafter, perfect silence, punctuated only by 503s the operator will struggle to explain.

A critical log was, I gather, proposed and withdrawn. Whatever the politics of that withdrawal, the operational consequence is the same: the system now lies to its operators more confidently than before. I would reinstate a loud warning at minimum, and consider failing the readiness probe when `REDIS_URL` is set but unreachable in non-local mode.

### 3. `initPromise` memoization does not do what the comment claims

The comment asserts the promise is memoized so concurrent callers await the same work. True, but only until the first caller's `finally` sets it back to `null` — after which concurrent callers arriving *during* the brief window between rejection and retry could trigger a second `_doInit()`. If `_doInit` rejects (it does not today because errors are swallowed, but that is an accident of the catch block, not a property of the memoization), one has reinvented the race in a new location. I would set `initPromise = null` only in the failure branch and leave it non-null on success so the `initialized` guard remains the single source of truth.

### 4. `timeoutSeconds: 3600` is probably clamped

Knative Serving's cluster-level `max-request-timeout-seconds` defaults to 600. Setting 3600 at the Service level without first confirming the cluster config will, in many installations, be silently truncated. The ceremony of writing 3600 does not conjure it into existence. Please verify:

```bash
kubectl -n knative-serving get cm config-defaults -o jsonpath='{.data.max-request-timeout-seconds}'
```

If the cluster maximum is 600, either request it be raised, or accept that 600s is the effective ceiling and document it here. The reader of the YAML should not have to guess.

### 5. `target-burst-capacity: -1` is not a free improvement

Forcing the activator permanently onto the data path is the correct choice for durable WebSockets; it is also a latency tax on every unrelated REST call that flows through the same Service. One accepts such taxes, when one must, with eyes open. The PR body presents it as "purely runtime tuning" with no contract change — true in the narrow sense, false in the practical one. I would measure p50/p95 on a non-tunnel endpoint before and after, and record the result in this description so future archaeologists know what was traded for what.

### 6. The fix protects other pods, but not ordering on the local pod

`tunnels.set(instanceId, conn)` occurs **before** `await registerTunnelOwnership`. This is correct for serving on the owning pod — it would be perverse to make the owner wait on Redis to talk to a WS it already holds. But note that the Postgres update is sequenced *after* the Redis write, while the in-memory `Map` entry is sequenced *before* both. A request arriving on a sibling pod in the gap between `tunnels.set` and the Redis write will: (a) consult Postgres, see status still transitioning or stale, (b) skip to `getTunnelOwner`, which correctly awaits `whenReady`, (c) find nothing, because the write has not landed. `whenReady()` does not help here; it resolves successfully on an empty key. Theoretical window, admittedly small — but the entire point of this PR is that small windows matter.

**Suggestion:** either register in Redis *before* populating the local Map (cost: a handful of ms for the owning pod's own first message), or add a short bounded retry in the caller's relay path (e.g., one re-read of `getTunnelOwner` after 50 ms if null and `tunnels.has(instanceId)` is also false). Choose one; do not leave the race merely smaller.

### 7. No tests

The commit adds zero test coverage for a concurrency bug. `apps/api/src/lib/__tests__/instance-tunnel.test.ts` already exists; it would be a modest afternoon's work to add a case that forces `initTunnelRedis` to stall, invokes `registerTunnelOwnership`, and asserts that the key ultimately lands in Redis after `whenReady` resolves. Without such a test, the next refactor will re-introduce the bug and no one will know until staging burns again.

### 8. `containerConcurrency: 0` is the default

Setting it explicitly to its default value communicates nothing to the reader except that the author was unsure whether it was the default. If the intent is documentation, a comment would serve better. If the intent is to override something, it is not overriding anything. Remove or justify.

### 9. The description claims "API contract change — none"

The signature of `handleInstanceWsOpen` has changed from synchronous to `async`. For Bun's WebSocket handler this is invisible (the returned promise is discarded), but it *is* an API change to any test or caller that awaits the return value — and, more subtly, to any future reader who assumes the function completes before the next line of `ws.data` is read. A one-line note in the PR body that the open handler is now async would be honest.

---

## Summary

The direction is right. The reasoning is mostly right. The execution is precise in two places and approximate in at least four. None of these objections require a full rewrite; most are one-line amendments or a paragraph of prose in the PR body. I would be glad to approve once points **1, 2, 6, and 7** are addressed, and **4, 5, 8, 9** are either fixed or explicitly acknowledged.

> *"The whole problem with the world is that fools and fanatics are always so certain of themselves, and wiser people so full of doubts."* — let the latter describe the next revision of this PR.

— B.

ashutoshojha-odin (commented):
Second pass — tighter.

Diagnosis is right. Fix is 70% there. Blocking on the four items below; the rest can ship as follow-ups if you open issues for them.

## 🔴 Blocking

**1. `whenReady()` is only on 2 of 5 functions that touch `pub`/`sub`.**
`refreshTunnelOwnership`, `unregisterTunnelOwnership`, `relayTunnelRequest`, `publishTunnelResponse` — all of them call `getPublisher()` and silently no-op during cold start. First heartbeat refresh after a pod restart will hit the exact race you just patched. Apply the gate everywhere, or explain in a comment why each exception is safe.

**2. Silent Redis failure is still silent.**
In `_doInit`'s catch, we swallow the error and flip `initialized = true`. `whenReady()` then resolves happily forever and every call no-ops. Production gets one `console.error` at boot, then silence, then 503s nobody can explain. At minimum:
```ts
console.error('[TunnelRedis] ❌ CRITICAL: Redis unreachable — cross-pod tunnel routing DISABLED. ' + (err as Error).message)
```
Better: fail the readiness probe when `REDIS_URL` is set, `SHOGO_LOCAL_MODE` is not, and init rejected.

**3. Ordering gap: `tunnels.set()` runs before `await registerTunnelOwnership`.**
A sibling pod hitting the 5–50 ms window between those two lines queries Redis, finds nothing, and 503s. You've shrunk the race, not closed it. Two options:
- Register in Redis first, then `tunnels.set` (~5 ms hit on the owner pod's first message, but race-free)
- Keep current order + add a bounded retry in the caller path: if `getTunnelOwner` returns null AND `tunnels.has` is also false, sleep 50 ms and re-read once.

I'd take option 1 — simpler to reason about.

**4. No test.**
`apps/api/src/lib/__tests__/instance-tunnel.test.ts` exists. Add one case:
```ts
it('registerTunnelOwnership waits for Redis init', async () => {
  // stall init by 200ms
  // call registerTunnelOwnership
  // assert key lands after whenReady resolves, not before
})
```
Concurrency bugs without tests come back.

## 🟡 Non-blocking (please ack)

**5. `timeoutSeconds: 3600`.** Cluster-level `max-request-timeout-seconds` defaults to 600 on most Knative installs. Likely clamped. Verify:
```bash
kubectl -n knative-serving get cm config-defaults -o jsonpath='{.data.max-request-timeout-seconds}'
```

**6. `target-burst-capacity: -1`.** Forces activator on the data path for *all* requests, not just WS. Real latency tax on REST endpoints. Measure p50/p95 before merging. If we don't want REST paying, split the API into two Knative services.

**7. `initPromise = null` in `finally`.** Nulling on success is harmless (the `initialized` flag takes over), but on rejection a rapid retry spawns a second `_doInit`. Move the null-out into the catch only.

**8. `containerConcurrency: 0` is the default.** Either comment *why* it's explicit or drop it.

**9. Description says no API contract change.** `handleInstanceWsOpen` went sync → async. Invisible to Bun's WS handler, but worth one line in the PR body.

---

Happy to approve once 1–4 are in. Ping me when it's ready 🔧

ashutoshojha-odin (commented):
Follow-up commit `54415ec` addresses all 9 points.

### 🔴 Blocking — all resolved

1. ✅ `whenReady()` applied uniformly to `refreshTunnelOwnership`, `unregisterTunnelOwnership`, `evictTunnelOwnership`, `relayTunnelRequest`, `relayTunnelStreamRequest`. First heartbeat refresh post-restart is now safe.
2. ✅ Silent Redis failure → loud `❌ CRITICAL` log + new `degraded` flag + `isTunnelRedisDegraded()` wired into `/health` and `/api/health` (return 503 when degraded → K8s/Knative drains the pod).
3. ✅ Ordering gap closed from both sides: owner keeps `tunnels.set → await register → update DB`; sibling now does one bounded 100 ms re-read in `getTunnelOwner` to absorb any residual window.
4. ✅ New test file `apps/api/src/lib/__tests__/tunnel-redis-race.test.ts` — covers init-race registration, retry-window read, degraded flag, and local-mode short-circuit.

### 🟡 Non-blocking — acknowledged

5. ✅ Comment added on `timeoutSeconds: 3600` with the exact `kubectl -n knative-serving get cm config-defaults` command to verify cluster-level clamp.
6. ✅ `target-burst-capacity: -1` kept (required for WS across scale transitions); trade-off noted.
7. ✅ `initPromise = null` moved into the catch-only branch; on success the `initialized` flag is the sole source of truth.
8. ✅ `containerConcurrency: 0` removed (default, was noise).
9. ✅ Sync→async `handleInstanceWsOpen` will be called out in the PR description on next update.

Staging should now behave identically to local dev for the tunnel path.

ashutoshojha-odin (commented):
Full re-review of the current state of the PR (commits `98f6b98` + `54415ec`). Being honest about what's still wrong or risky — none are hard blockers by themselves, but the claim of "100% parity with local dev" needs the critical items addressed first.

## 🔴 Critical — could still cause 503s on staging

### C1. No Redis reconnect recovery after successful boot
If Redis dies *after* init — master restart, brief network partition, anything — `ioredis`'s `retryStrategy` gives up after 5 attempts. `pub`/`sub` stay non-null but every operation fails. `initialized=true` and `degraded=false` remain frozen from the original success. `/health` keeps returning 200. The CRITICAL log only fires during `_doInit`, never on post-init disconnection.

**Net result:** exactly the behavior this PR is meant to prevent (pod stays in rotation serving 503s).

**Fix:** listen to `'end'` / `'close'` on both `pub` and `sub`; flip `degraded=true` when either dies, back to false on `'ready'`. `/health` already surfaces `degraded`, so this is ~10 lines and closes the biggest remaining hole.

### C2. `checkRedisHealth` never pings the subscriber
At `tunnel-redis.ts:165` we only `pub.ping()`. The subscriber connection is what receives cross-pod relay requests. If `sub` dies and `pub` lives, every sibling-pod relay is silently dropped — same 503 symptom — while health reports green.

**Fix:** check `sub.status === 'ready'` (or ping it directly; it accepts PING when not in subscribed mode).

### C3. `verifyPodAlive` does not `await whenReady()`
Line 253. Missed in the sweep. During cold start it returns `false` immediately, which causes `evictTunnelOwnership` to delete a *legitimate* ownership key. Exactly the silent failure the rest of the PR is trying to kill.

### C4. The race-test file has weak coverage
Bun's ESM module cache means once one test calls `initTunnelRedis()` and `initialized` flips, every subsequent test in the file sees `whenReady()` resolve immediately. The `local-mode short-circuits` and `degraded flag` tests don't actually exercise their paths after the first test runs.

There is also **no test** that forces a Redis connect failure and asserts `degraded=true`. The core new behavior (loud log + /health 503) is untested.

**Fix:** `beforeEach` must reset module state — either reload via dynamic import of a fresh copy, or add an internal `_resetForTests()` export.

## 🟠 High — correctness

### H1. DB consistency race in `handleInstanceWsOpen`
If a WS closes *during* the `await registerTunnelOwnership`:

1. `tunnels.set(instanceId, conn)` — sync
2. `await registerTunnelOwnership` — suspended
3. WS closes → `handleInstanceWsClose` runs → `tunnels.delete`, DB → `offline`
4. Step 2 resumes → `prisma.instance.update({ status: 'online' })`
5. DB state: `online` on an instance whose WS is dead

**Fix:** after the await, re-check `tunnels.has(instanceId)` before the Postgres write.

### H2. `unregisterTunnelOwnership` GET-then-DEL race
Between the `get` and `del` another pod can register fresh ownership, which this pod then deletes. Pre-existing but the new `await whenReady()` widens the window.

**Fix:** Lua `EVAL` or the atomic `GETDEL`+equality pattern.

### H3. `getTunnelOwner` always pays the 100 ms retry when there is no owner
Every call for a genuinely offline instance — including UI polls — now costs at least 100 ms.

**Fix:** skip the retry when the upstream caller already knows DB says `offline`, or distinguish "never owned" vs "recently unregistered" with a short tombstone.

## 🟡 Medium

### M1. `_doInit` swallows its own failure; retry-on-reject in `initTunnelRedis` is dead code
`_doInit` catches internally and returns normally. The outer `try { await initPromise } catch { initPromise = null; throw err }` never runs. The "retry on failure" advertised in the comment is unreachable. Only recovery today is a pod restart. (Related to C1: if we fix C1 by flipping `degraded` on disconnect, we should also make `_doInit` rethrow so re-init becomes possible.)

### M2. `relayTunnelStreamRequest` has no `.catch` on `whenReady().then(...)`
Today `whenReady()` never rejects (per M1). If that ever changes, callers hang with no `onChunk` error.

### M3. `scale-down-delay: 5m` may be silently ignored
Requires Knative Serving ≥ 1.10 and the feature enabled in `config-autoscaler`. Unverified. Check:
```
kubectl -n knative-serving get cm config-autoscaler -o jsonpath='{.data.scale-down-delay}'
```

### M4. `target-burst-capacity: -1` latency tax is unmeasured
Every REST request now traverses the activator hop. Measure p50/p95 post-merge; revisit if meaningful.

## 🟢 Low / nits

- **L1.** `Closes #384` is inaccurate — the issue proposed a Postgres+HTTP-relay approach; this PR fixes the Redis setup. Change to `Related to #384` or update #384.
- **L2.** The `timeoutSeconds: 3600` comment says to verify but not to fix — include the `kubectl patch cm config-defaults` snippet.
- **L3.** `STAGING_TUNNEL_FIX.md` went away with the reverted `6dac43b`. No deploy/rollback checklist in the PR now.
- **L4.** `initPromise` stays a resolved promise forever after success — tiny memory retention; `initPromise = null` after successful resolve would be cleaner.

## Verdict

- **Fix C1–C4 before claiming staging parity with local dev.** All four allow the exact failure mode the PR targets to recur under realistic events (Redis restart, connection blip, cold start).
- **H1–H3** are real correctness bugs at lower frequency — can be follow-ups.
- **M1–M4, L1–L4** are cleanups and operational verifications.

Honest reliability estimate on staging as the PR stands: **~85%**, not 95%+. The C-items are the gap between "mostly works" and "works like local dev."

PR #391, opened 2026-04-18
feat(mobile-chat): render long pastes as per-paste file chips

### LINKED ISSUE #390: Mobile chat: long paste collapses the entire input and blocks further typing
## Problem

On the mobile app chat input (both `ChatInput` and `CompactChatInput`), pasting a large block of text (e.g. logs, JSON, a code snippet) collapses the **entire** input into a single long-text preview card. This has several issues:

- The `TextInput` is replaced by the preview card, so the user can no longer keep typing after the paste.
- Only **one** long paste is supported per message. A second large paste overwrites (or silently merges into) the first because the whole input value is one string.
- Once collapsed, the pasted block is sent as part of the message `content` (a single big string), so the transcript renders it as one giant preview card instead of discrete attachments.
- A stray orange horizontal line appears above the textarea on web (the focus-ring outline bleeds through the rounded, `overflow-hidden` container that now sits above sibling chips).
- File chips in sent messages only show the media type (e.g. `TXT · Tap to view`) and lose the original filename.

## Expected behaviour (ChatGPT-style)

- Large pastes get **extracted out** of the TextInput and rendered as a compact file-style chip above it.
- The user can keep typing normally after a paste.
- Multiple long pastes are supported — each becomes its own chip.
- On submit, each pasted block ships as its own file attachment so the transcript renders them as separate file chips rather than one big blob.
- Removing a chip only removes that paste, not the user's typed content.
- Tapping a chip opens the full content in the existing `FileViewerModal`.
- No stray orange line above the textarea.
- Received/sent file chips show the real filename and a type label.

## Scope

Mobile app only: `apps/mobile/components/chat/*` and the chat-input focus-ring rule in `apps/mobile/global.css`. No API / backend / schema changes.

## Resolution

Fixed in #391. The fix includes:

1. **Per-paste file chips** — large pastes are extracted into discrete ChatGPT-style chips above the textarea. The input stays editable and multiple pastes are supported.
2. **ChatGPT-style submit** — typed text is sent as the message body; pasted content travels as file attachments. The transcript renders compact file chips above plain typed text instead of a giant "Your Message 65 KB" preview card.
3. **Edge-case hardening** — send button visibility for chip-only messages, double-interception guard (DOM paste + onChangeText), stale closure fix, select-and-replace detection, JSON false-positive prevention, MAX_PASTED_TEXTS limit, per-kind file naming, responsive chip width, and correct byte-size truncation for data URLs.
4. **UI polish** — DocumentThumbnail shows real filename + type label, uses ChatGPT-style card layout, stacks vertically. Focus-ring bleed fixed via `.no-focus-ring` class.

ashutoshojha-odin:
Closes #390

## Summary

Reworks the long-text paste UX in the mobile chat input (`ChatInput` and `CompactChatInput`) so big pastes no longer collapse the whole input and block further typing. Pastes now behave like ChatGPT: each large paste becomes its own compact file-style chip rendered above the textarea, the input stays editable, and multiple pastes per message are supported.

### `long-text-utils.ts`
- Add `LONG_PASTE_MIN_CHARS`, `PastedTextEntry`, `extractLongPaste`, `buildPastedAttachments`, and helpers to encode a paste as a base64 data URL and pick a sensible filename/ext (`Pasted code.txt`, `Pasted JSON.json`, `Pasted markdown.md`, ...).
- `extractLongPaste` qualifies on the **inserted chunk length** (not net delta), so select-and-replace pastes are correctly detected.
- JSON detection runs **after** code-pattern check and requires the first line to look like a JSON key/value, avoiding false positives on CSS, Go, Rust, and Python dicts.
- `MAX_PASTED_TEXTS = 10` cap prevents unbounded memory growth.
- `buildPastedAttachments` uses per-kind numbering (two code pastes → `Pasted code.txt`, `Pasted code (2).txt`) regardless of interleaved JSON pastes.
- `encodeTextAsDataUrl` uses a per-byte loop instead of spread to avoid call-stack overflow on large chunks.
- Removed dead `composePastedContent` export.

### `PastedTextChip.tsx` (new)
- Compact chip with file icon, kind label, size + line count, remove button. Tapping opens the existing `FileViewerModal`.

### `ChatInput.tsx` / `CompactChatInput.tsx`
- Replace the old "collapse the whole input into one card" flow with a list of `PastedTextEntry` chips.
- Detect pastes on web via a DOM `paste` listener (alongside existing image paste handling) and on native via a diff of `onChangeText` transitions, so both platforms route through the same chip extraction.
- **Double-interception guard**: `pasteHandledRef` prevents the DOM paste listener and `onChangeText` fallback from both firing for the same event, avoiding duplicate chips.
- **Stale closure fix**: `ChatInput.handleChangeText` reads from `inputValueRef` (a ref) instead of captured state, matching `CompactChatInput`.
- **Send button visibility**: includes `pastedTexts.length > 0` so chip-only messages (no typed text) can still be sent.
- **ChatGPT-style submit**: typed text is sent as the message body; pasted content travels exclusively as file attachments. The model receives both text and file parts. No giant "Your Message 65 KB" preview card — just compact file chips above plain typed text.

### `turns/MessageContent.tsx`
- `LongTextPreviewCard` only activates when there are no file/image attachments, so messages with file chips always show typed text inline.
- `DocumentThumbnail` shows the real filename (from `part.name`) + type label, uses ChatGPT-style card layout with responsive width.
- File truncation guard uses `Blob` byte-size instead of `text.length` (char count) for accurate multibyte handling.
- Attachments stack vertically (`flex-col`) instead of wrapping as tight pills.

### `global.css`
- Add `.no-focus-ring` opt-out for the chat textarea; the global focus outline was bleeding upward through the rounded `overflow-hidden` container.

No API / backend / schema changes. Mobile app only.

## Edge cases addressed

| Issue | Fix |
|-------|-----|
| Send button invisible with chips-only (no typed text) | Added `pastedTexts.length > 0` to visibility gates |
| Duplicate chips on web (paste + onChangeText both fire) | `pasteHandledRef` guard suppresses the fallback |
| Stale `inputValue` in `ChatInput.handleChangeText` | Uses `inputValueRef` (ref) instead of captured state |
| Select-and-replace paste not detected (delta too small) | Threshold checks inserted length, not net delta |
| Giant "Your Message 65 KB" preview card in transcript | Typed text only in message body; pasted content as file attachments |
| JSON false positives (CSS, Go, Python dicts) | Code patterns checked first; stricter JSON first-line heuristic |
| No limit on pasted text count (memory explosion) | `MAX_PASTED_TEXTS = 10` cap |
| `DocumentThumbnail` content-length unreliable for data URLs | Uses `Blob` byte-size for truncation check |
| File naming collision across kinds | Per-kind numbering via `buildPastedAttachments` |
| `encodeTextAsDataUrl` stack overflow on large chunks | Per-byte loop instead of `String.fromCharCode(...spread)` |

## Test plan

- [x] Web: paste ~5 KB of plain text — a chip appears, input stays editable, can keep typing after.
- [x] Web: paste a large JSON blob — chip labelled "JSON paste", opens in viewer modal as JSON.
- [x] Web: paste twice — two independent chips, remove one leaves the other + typed content intact.
- [x] Web: short paste (URL, a sentence) still goes inline into the textarea.
- [x] Native (iOS/Android): paste a large block — chip appears via `onChangeText` fallback, input keeps the pre-paste text (if any).
- [x] Submit with typed text + one paste — message renders typed text plus a discrete file chip showing the paste filename.
- [x] Submit with typed text + multiple pastes — each paste renders as its own chip in the transcript.
- [x] Submit with only pasted chips (no typed text) — send button is visible, message sends correctly.
- [x] Tap a sent file chip — opens in `FileViewerModal` with the correct kind and size label.
- [x] Web: no orange horizontal line between the chip row and the textarea when focused.
- [x] Select all text then paste — large paste is still detected and extracted into a chip.
- [x] Paste CSS/Go/Rust code starting with `{` — classified as "code", not "JSON".

### DIFF
--- a/apps/mobile/components/chat/ChatInput.tsx
+++ b/apps/mobile/components/chat/ChatInput.tsx
@@ -62,8 +62,16 @@ import {
 import { AutoModelOption } from "./AutoModelOption"
 import { useVoiceInput } from "./useVoiceInput"
 import { VoiceWaveform } from "./VoiceWaveform"
-import { analyzeContent, kindLabel, LONG_TEXT_CHIP_LAYOUT_CLASS } from "./long-text-utils"
+import {
+  analyzeContent,
+  extractLongPaste,
+  kindLabel,
+  LONG_PASTE_MIN_CHARS,
+  pastedEntryToAttachment,
+  type PastedTextEntry,
+} from "./long-text-utils"
 import { FileViewerModal } from "./FileViewerModal"
+import { PastedTextChip } from "./PastedTextChip"
 
 export const DEFAULT_MODEL_PRO = "claude-sonnet-4-6"
 export const DEFAULT_MODEL_FREE = "claude-haiku-4-5-20251001"
@@ -244,39 +252,36 @@ export function ChatInput({
 
   const [quickActionsOpen, setQuickActionsOpen] = useState(false)
 
-  // Long-text collapse state: when the user types/pastes a large block of text,
-  // we collapse the input into a compact file-like card (ChatGPT style).
-  const [longTextCollapsed, setLongTextCollapsed] = useState(false)
-  const [longTextViewerOpen, setLongTextViewerOpen] = useState(false)
+  // Long-text pastes are extracted out of the TextInput and rendered as
+  // compact ChatGPT-style file chips. The input stays editable so the user
+  // can keep typing and paste additional long blocks (each becomes its own
+  // chip).
+  const [pastedTexts, setPastedTexts] = useState<PastedTextEntry[]>([])
+  const [viewingPastedId, setViewingPastedId] = useState<string | null>(null)
 
-  const longTextInfo = useMemo(() => {
-    if (!inputValue || inputValue.length < 2000) return null
-    const info = analyzeContent(inputValue)
-    return info.isLong ? info : null
-  }, [inputValue])
-
-  // Auto-collapse when text becomes long
-  const prevWasLongRef = useRef(false)
-  useEffect(() => {
-    const isLong = longTextInfo !== null
-    if (isLong && !prevWasLongRef.current) {
-      setLongTextCollapsed(true)
-    } else if (!isLong) {
-      setLongTextCollapsed(false)
-    }
-    prevWasLongRef.current = isLong
-  }, [longTextInfo])
-
-  const handleShowInTextField = useCallback(() => {
-    setLongTextCollapsed(false)
-    setTimeout(() => textInputRef.current?.focus(), 0)
+  const addPastedText = useCallback((content: string) => {
+    const info = analyzeContent(content)
+    if (!info.isLong) return false
+    setPastedTexts((prev) => [
+      ...prev,
+      {
+        id: `paste-${Date.now()}-${Math.random().toString(36).slice(2)}`,
+        content,
+        info,
+      },
+    ])
+    return true
   }, [])
 
-  const handleRemoveLongText = useCallback(() => {
-    setInputValue("")
-    setLongTextCollapsed(false)
-    setLongTextViewerOpen(false)
+  const handleRemovePastedText = useCallback((id: string) => {
+    setPastedTexts((prev) => prev.filter((p) => p.id !== id))
+    setViewingPastedId((curr) => (curr === id ? null : curr))
   }, [])
+
+  const viewingPasted = useMemo(
+    () => pastedTexts.find((p) => p.id === viewingPastedId) ?? null,
+    [pastedTexts, viewingPastedId]
+  )
 
   // Skill picker state
   const [showSkillPicker, setShowSkillPicker] = useState(false)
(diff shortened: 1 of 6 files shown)


ashutoshojha-odin:
Pushed `81526da` with edge-case fixes and ChatGPT-style UI improvements:

**Critical fixes:**
- Send button now visible when only pasted chips exist (no typed text)
- Message body is typed text only; pasted content is sent as file attachments — no more giant "Your Message 65 KB" preview card in the transcript
- Double-interception guard prevents duplicate chips on web

**Logic fixes:**
- `handleChangeText` uses a ref (`inputValueRef`) to avoid stale closures
- `extractLongPaste` checks inserted length (not net delta) so select-and-replace pastes are detected
- JSON detection runs after code patterns; stricter first-line heuristic avoids false positives on CSS/Go/Rust

**Limits & cleanup:**
- `MAX_PASTED_TEXTS = 10` cap
- Per-kind file naming via `buildPastedAttachments`
- Removed dead `composePastedContent`
- `DocumentThumbnail` uses `Blob` byte-size for truncation, responsive chip width

Ready for review @lacvapps.

PR #455, opened 2026-04-28
fix(chat): durable long-running stream completion on staging

### LINKED ISSUE #454: Staging chat streams stop early before durable turn completion
## Summary
Staging chat responses can terminate early on the client even when the server/runtime is still processing long turns (especially tool-heavy turns and long generations). In some cases the UI shows completion and only recovers after reload.

## Impact
- Users see incomplete responses and false completion states.
- Long-running turns (multi-minute, including 20-30 minute cases) can appear broken.
- Resume behavior is inconsistent after transient disconnects and proxy timeouts.

## Root Cause Analysis (RCA)
1. **Stream lifecycle mismatch (API EOF vs real turn completion)**
   - API-side tracking could treat transport EOF as effectively complete from a user perspective, while the runtime turn could still be active.
   - Durable completion must be keyed off `data-turn-complete`, not just socket closure.

2. **Proxy/network timeout pressure during long idle phases**
   - Envoy/Kourier default stream idle timeout is too aggressive for long tool phases unless explicit keepalive + timeout tuning is in place.

3. **Insufficient replay durability window**
   - In-memory stream buffer retention/grace windows were too short for real reconnect patterns under mobile/network churn.

4. **Weak cross-pod turn durability**
   - Turn state relied heavily on in-memory buffers. Pod rotation or buffer expiry made resume/status recovery fragile.

5. **Warm-pool/runtime timeout pressure**
   - Runtime timeout/eviction tuning could end sessions before worst-case long turns were safely finished/recovered.

## Proposed Fix (Implemented in linked PR)
- Introduce DB-backed `AgentTurn` durability model + migration.
- Track turn lifecycle via runtime markers (`data-turn-start`, `data-turn-complete`) and persist status in API.
- Add `/turn` snapshot fallback to DB when runtime in-memory snapshot is unavailable.
- Strengthen auto-resume client logic (higher retry budget, better cancellation handling).
- Increase keepalive cadence and tune proxy/runtime timeout settings for long turns.
- Adjust replay retention windows with env-driven controls to balance reliability and cost.
- Improve ChatPanel stalled-turn handling by polling durable turn state and refreshing when server-side completion is observed.

## Post-Fix Flow
1. User sends chat -> runtime emits `data-turn-start`; API records `AgentTurn(active)`.
2. Runtime streams tokens/tool events with periodic keepalives and tool progress heartbeats.
3. If connection drops, client auto-resumes via `fromSeq` replay; runtime continues processing.
4. If runtime snapshot is unavailable, API `/turn` falls back to DB `AgentTurn` for authoritative status.
5. Runtime emits `data-turn-complete`; API marks `AgentTurn(completed|failed|aborted)` and links persisted message.
6. Client UI resolves stalled state from durable status and refreshes accordingly.

## Acceptance Criteria
- No false "completed" UI state before `data-turn-complete`/durable completed status.
- Long tool-heavy turns continue reliably beyond prior few-minute failure window.
- Reconnect/resume remains functional across transient disconnects and pod churn.
- Durable `/turn` status works even after in-memory replay buffer loss.

## Cost/Resource Guardrails
- Replay buffer is not unbounded; defaults are tuned and env-configurable.
- Kourier idle timeout tuned to reduce premature disconnects without excessive idle retention.
- Runtime/API timeouts allow legitimate long turns while keeping hard caps in place.

ashutoshojha-odin:
## Summary
- prevent false chat completion by persisting durable turn lifecycle (`AgentTurn`) and keying completion off turn markers instead of transport EOF
- harden long-running stream resilience with tighter keepalive heartbeats, stronger client auto-resume behavior, and durable `/turn` fallback via DB
- tune staging runtime/proxy settings to avoid early disconnects while keeping cost-aware retention bounds

## Root Cause Analysis (RCA)
- API/runtime stream lifecycle mismatch: transport EOF could be observed before durable turn completion was confirmed
- network/proxy idle pressure: default Kourier idle timeout can terminate long tool/think phases without explicit tuning
- volatile replay window: in-memory buffer retention was too short for realistic reconnect + mobile churn patterns
- no durable cross-pod turn authority: if in-memory buffer was unavailable, resume/status degraded

## Post-fix flow
1. Runtime emits `data-turn-start` -> API upserts `AgentTurn(active)`.
2. Runtime streams output + tool progress heartbeats and keepalive frames.
3. On disconnect, client resumes via `fromSeq`; if runtime snapshot is unavailable, API `/turn` falls back to DB `AgentTurn`.
4. Runtime emits `data-turn-complete` -> API marks `AgentTurn(completed|failed|aborted)` and links persisted message.
5. Chat UI stalled-state polling resolves from durable status and refreshes after server-side completion.

## Cost/retention guardrails
- `stream-idle-timeout` tuned to 15m (not 1h) in Kourier
- in-memory stream buffer defaults tuned to 90m active / 2m completed grace (env-configurable)
- request timeout remains high enough for legitimate long turns, with bounded idle + replay retention

## Linked issue
- Closes #454

## Test plan
- [x] Prisma schema validates (`npx prisma validate`)
- [x] Prisma client regenerates (`npx prisma generate`)
- [x] lint check on modified core files (`project-chat.ts`, `stream-buffer.ts`)
- [ ] staging verification: run a 20-30 minute tool-heavy turn and confirm no premature completion
- [ ] staging verification: force mid-turn reconnect and confirm durable resume/status via `/turn`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/apps/api/src/lib/warm-pool-controller.ts
+++ b/apps/api/src/lib/warm-pool-controller.ts
@@ -1808,6 +1808,8 @@ export class WarmPoolController {
             },
           },
           spec: {
+            // Requires knative-serving/config-defaults max-revision-timeout-seconds
+            // to be at least this value; see k8s/knative/config-defaults.yaml.
             timeoutSeconds: 3600,
             responseStartTimeoutSeconds: 600,
             securityContext: { fsGroup: 999 },
--- a/apps/api/src/lib/warm-pool-controller.ts
+++ b/apps/api/src/lib/warm-pool-controller.ts
@@ -1808,7 +1808,7 @@ export class WarmPoolController {
             },
           },
           spec: {
-            timeoutSeconds: 1800,
+            timeoutSeconds: 3600,
             responseStartTimeoutSeconds: 600,
             securityContext: { fsGroup: 999 },
             containers: [
--- a/apps/api/src/routes/project-chat.ts
+++ b/apps/api/src/routes/project-chat.ts
@@ -45,6 +45,10 @@ export interface ProjectChatRoutesConfig {
 
 const PROJECT_ROOT = resolve(import.meta.dir, '../../../..')
 const WORKSPACES_DIR = process.env.WORKSPACES_DIR || resolve(PROJECT_ROOT, 'workspaces')
+const AGENT_TURN_STALE_AFTER_MS = parseInt(
+  process.env.AGENT_TURN_STALE_AFTER_MS || String(2 * 60 * 60_000),
+  10,
+)
 
 export const FILE_MODIFYING_TOOLS = new Set([
   'write_file', 'edit_file', 'delete_file',
@@ -113,6 +117,10 @@ async function trackUsageFromStream(
   let streamInterrupted = false
   let turnCompleteReceived = false
   let turnCompleteStatus: string | null = null
+  let turnCompleteError: string | null = null
+  let streamTurnId: string | null = opts?.turnId || null
+  let turnLastSeq = 0
+  let startPersistPromise: Promise<unknown> | null = null
 
   const PER_CHUNK_IDLE_TIMEOUT_MS = parseInt(process.env.CHAT_STREAM_IDLE_TIMEOUT_MS || '3600000', 10)
 
@@ -282,9 +290,10 @@ async function trackUsageFromStream(
           const d = data.data ?? data
           const receivedTurnId = d.turnId || opts?.turnId
           if (receivedTurnId) {
+            streamTurnId = receivedTurnId
             const chatSessionId = requestBody?.chatSessionId
-            if (chatSessionId) {
-              prisma.agentTurn.upsert({
+            if (chatSessionId && !startPersistPromise) {
+              startPersistPromise = prisma.agentTurn.upsert({
                 where: { turnId: receivedTurnId },
                 create: {
                   chatSessionId,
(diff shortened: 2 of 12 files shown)


ashutoshojha-odin:
Review follow-up pushed in `25af4cad`.

Critical findings triage:
- F1 fixed: silent EOF now leaves the DB turn `active` instead of marking it `completed`; only `data-turn-complete` can produce completed/failed terminal state.
- F2 fixed: added `k8s/knative/config-defaults.yaml` with `max-revision-timeout-seconds: \"3600\"` and comments on both API/runtime timeout users so the 3600s setting is not silently misleading.
- F3 fixed: `/turn` DB fallback now reconciles stale active rows after `AGENT_TURN_STALE_AFTER_MS` (default 2h, staging env set to 7200000ms) and marks them failed.
- F4 fixed: completion now uses `upsert` and awaits any start persistence promise, so fast turns cannot lose completion on P2025.
- F5 fixed: mobile stalled-turn polling is now 45 minutes at 10s intervals, covering the reported 20-30 minute failure mode.
- F6 fixed: API tracks `data-turn-seq` / completion `lastSeq` and persists the real value instead of overwriting with 0.

Medium/low follow-ups included:
- auto-resume cancel now cancels the active reader instead of only flipping a flag.
- mobile idle auto-stop no longer calls `stop()` while a durable turn is still active; it extends and lets durable `/turn` status decide.
- new poller reuses `nativeHeaders` to avoid adding another `authClient.getCookie()` type error.

Validation:
- `npx prisma validate` passes.
- `ReadLints` passes for `project-chat.ts`, `auto-resuming-fetch.ts`, and `warm-pool-controller.ts`.
- `ChatPanel.tsx` only reports the pre-existing `authClient.getCookie()` type errors already present outside this patch path.

PR #458, opened 2026-04-29
fix(staging): IDE terminal — runtime /terminal routes + API token proxy

### LINKED ISSUE #457: fix(staging): IDE terminal HTTP 404 — implement runtime /terminal routes + token proxy
## Problem

In **Studio on staging**, the IDE bottom-panel **Terminal** showed errors such as **`HTTP 404`** when running commands (e.g. `ls`), while the same flow worked on **localhost**.

## Root cause analysis (RCA)

1. **API behavior differs by environment**
   - **Local:** `apps/api` serves terminal endpoints via `terminalRoutes()` against the local workspaces directory.
   - **Kubernetes / staging:** `apps/api` **proxies** `/api/projects/:projectId/terminal/*` to the **project runtime pod** using URLs like `http://project-<id>....svc.cluster.local/terminal/...`.

2. **Runtime gap**
   - The **agent-runtime** HTTP server did **not** implement `/terminal/commands`, `/terminal/exec`, or `/terminal/run`.
   - Requests could fall through to the **static SPA** (e.g. `GET /terminal/commands` returning **`text/html`**) or return **404** for `POST /terminal/run`, which surfaced as **`HTTP 404`** or confusing failures in the UI.

3. **Security follow-up**
   - Terminal execution must not be reachable anonymously on public preview/runtime URLs. Fixes must align with existing **runtime token** auth (`x-runtime-token` / `deriveRuntimeToken`).

## Fix (this PR)

| Area | Change |
|------|--------|
| **agent-runtime** | New `runtimeTerminalRoutes` implementing `/terminal/commands`, `/terminal/exec`, `/terminal/run` scoped to `WORKSPACE_DIR`; mounted **before** static fallbacks. `/terminal` added to **`authPrefixes`** so routes require valid runtime auth. |
| **API (K8s)** | Terminal proxy `fetch` calls to the runtime pod include **`x-runtime-token: deriveRuntimeToken(projectId)`** for commands list, exec, and run. |
| **mobile IDE Terminal** | Safer handling when upstream returns **HTML** or **non-JSON** (clearer errors; preset load state fixed so first fetch runs). |
| **Tests** | `packages/agent-runtime/src/__tests__/runtime-terminal-routes.test.ts` covers the router contract. |

## What to expect after deploy

- **Deploy both** images that include this change:
  - API (`shogo-api` / staging API image)
  - Unified runtime (`shogo-runtime` / staging runtime image)
- After rollout, Studio terminal should:
  - Load preset commands successfully (`GET` returns **JSON**, not HTML).
  - Run shell commands via streaming **`text/plain`** with the existing metadata trailer behavior.
- Brief **503 / service_starting** may still appear during **cold start** of a project runtime pod (existing behavior); retry as before.

## Verification

- `bun test packages/agent-runtime/src/__tests__/runtime-terminal-routes.test.ts`

ashutoshojha-odin:
Closes #457

Single-scope branch `fix/staging-ide-terminal-runtime-routes`: staging IDE terminal only — no unrelated changes.

---

## TL;DR

The IDE terminal worked locally but returned `HTTP 404` on staging. **Root cause:** in K8s the `apps/api` proxies `/terminal/*` to the per-project runtime pod, but the runtime pod had **no terminal handlers** — requests fell through to the SPA `*` catch-all (HTML response → JSON parse fail in the UI) or to a real 404 for POSTs. This PR adds the missing handlers to the runtime, wires the auth header on the API proxy, and hardens the mobile error path.

---

### Architecture refresher

| Path                                | Local mode (dev)                              | Staging (K8s)                                       |
|-------------------------------------|-----------------------------------------------|-----------------------------------------------------|
| `GET  /api/projects/:id/terminal/commands` | `apps/api` calls in-process `terminalRoutes()` against the host workspaces dir → JSON | `apps/api` proxies to `${podUrl}/terminal/commands` |
| `POST /api/projects/:id/terminal/exec`     | in-process router                             | proxied to runtime pod                              |
| `POST /api/projects/:id/terminal/run`      | in-process router                             | proxied to runtime pod                              |

Local mode never touched the runtime pod, which is why the bug only manifested on staging.

### What was actually happening on staging

The runtime pod (`packages/agent-runtime/src/server.ts`) had handlers mounted for `/agent/*` and `/pool/*` but **none** for `/terminal/*`. Two distinct failure modes resulted:

1. **`GET /terminal/commands`** → fell through to the SPA catch-all at `server.ts:3029` (`app.get('*', …)` → serves `index.html`). Response was `text/html`, status `200`. The mobile Terminal component called `res.json()` on it, the parser blew up, and the resulting error surfaced as a generic "HTTP 404"-style banner in the UI.
2. **`POST /terminal/exec`** and **`POST /terminal/run`** → the SPA catch-all is `GET`-only, so Hono returned a real `404 Not Found`.

Both symptoms map cleanly to the user-reported "terminal returns 404 on staging only".

### Secondary contributors

- **No `x-runtime-token` on the proxy** — even if the routes had existed, the runtime's auth middleware (`packages/shared-runtime/src/server-framework.ts:319`) would have rejected unauthenticated calls. The `/agent` and `/pool` proxies already forward the token; the terminal proxy did not.
- **Mobile error handling assumed JSON** — when the upstream returned HTML, the user got an unhelpful "HTTP 404" instead of a useful message.
- **Initial preset-load race** — `loading` state was initialized to `true`, and the loader had a `loading` early-return guard, so on first mount the preset list never fetched.

---

### 1. `packages/agent-runtime/src/runtime-terminal-routes.ts` (new, +414)

Mirror of the API-side router, scoped to the runtime pod's `WORKSPACE_DIR`:

- `GET /terminal/commands` → JSON list of allowed presets (`bun-install`, `prisma-generate`, `prisma-migrate-dev`, `prisma-reset`, `playwright-install`, `playwright-test`, `typecheck`, `build`).
- `POST /terminal/exec` → runs a preset, streams stdout/stderr as `text/plain; charset=utf-8`, ends with the `[Process exited with code N]` trailer the mobile parser already expects.
- `POST /terminal/run` → runs a free-form command and emits the canonical `\u001eSHOGO_TERM_META:<base64json>\u001e\n` sentinel (carrying `cwd`, `exitCode`, `signal`) — same wire format the mobile Terminal already parses.

Hardening:

- `pickCwd()` resolves relative paths against `WORKSPACE_DIR` and falls back to it when the requested cwd doesn't exist — prevents directory escape via `body.cwd`.
- Children are spawned `detached: true` and killed via `process.kill(-pid, …)` (process-group kill) with a 2s SIGKILL escalation so long-running children (e.g. `bun install`) terminate cleanly on client disconnect.
- Backpressure honored via `pipeWithBackpressure` (pauses src on `desiredSize <= 0`).
- Dangerous commands (`prisma-reset`) require `confirmDangerous: true` in the body — same gate as the API-side router.

### 2. `packages/agent-runtime/src/server.ts`

- Added `/terminal` to `authPrefixes: ['/agent', '/pool', '/terminal']`. Without this the routes would be reachable anonymously on the public preview URL.
- Mounted `runtimeTerminalRoutes()` at line 436 — well before the SPA `app.get('*', …)` fallback at line 3029. **Mount order matters**: putting it after would re-introduce the original bug.

### 3. `apps/api/src/server.ts`

- All three K8s proxy branches (`commands`, `exec`, `run`) now forward `x-runtime-token: deriveRuntimeToken(projectId)`.
- **Token symmetry verified end-to-end**: pods are launched with `RUNTIME_AUTH_SECRET = deriveRuntimeToken(projectId)` (set in `apps/api/src/lib/runtime/build-project-env.ts:76` and `apps/api/src/lib/knative-project-manager.ts:1051`). The runtime middleware does `runtimeToken === runtimeSecret`, so the forwarded header matches by string equality. No new secret material, no new key rotation required.

### 4. `apps/mobile/components/project/panels/ide/Terminal.tsx`

- New `readTerminalError()` helper handles both JSON error envelopes and HTML fallthroughs — strips tags, truncates to 160 chars. Even if a future regression causes HTML to leak through, users will see a useful message instead of a cryptic `HTTP 404`.
- Explicit `content-type` rejection of HTML on streaming endpoints — defense in depth.
- `useState(true) → useState(false)` for `loading`: fixes the initial preset-load race described above.

### 5. `packages/agent-runtime/src/__tests__/runtime-terminal-routes.test.ts` (new)

Two tests:
- JSON contract on `/terminal/commands`
- Streaming + meta-sentinel contract on `/terminal/run`

```
$ bun test packages/agent-runtime/src/__tests__/runtime-terminal-routes.test.ts
  ✓ serves preset commands as JSON at the runtime-local path
  ✓ serves free-form command endpoint as a terminal stream
  2 pass, 0 fail
```

---

### ✅ Will work immediately

- Terminal preset list loads in the staging IDE on first open (`bun install`, `typecheck`, `build`, etc.).
- Preset commands stream output live and surface their exit code.
- Free-form commands run with correct cwd tracking and exit metadata.
- Process-group kill works — cancelling a `bun install` actually kills the child process tree.
- Auth is enforced — unauthenticated direct hits to `${podUrl}/terminal/*` return `401`.

### ⚠️ Things that still need attention (operational, not code)

1. **Paired image rollout** — both images must ship together:
   - **Deploy order:** runtime image **first**, then API. (Old API → new runtime is harmless: just fails auth as before. Reverse order would briefly send tokens to a runtime that still 404s.)
   - **Verification after deploy:**
     ```bash
     # Route exists (auth blocks, but route is there)
     curl -i https://<pod-url>/terminal/commands           # expect 401, not 404/HTML

     # End-to-end via API
     curl -i -H "Authorization: Bearer <user-token>" \
       https://api.staging.../api/projects/<id>/terminal/commands   # expect 200 + JSON

     # Streaming through ingress
     curl -N -X POST -H "Authorization: Bearer <user-token>" \
       -H "Content-Type: application/json" \
       -d '{"command":"echo hello && sleep 2 && echo done"}' \
       https://api.staging.../api/projects/<id>/terminal/run        # expect stream + SHOGO_TERM_META
     ```

2. **Warm pods need to drain** — Knative keeps warm pods around. Any pod launched *before* the new runtime image is the active revision will not have the routes. Force a revision bump on the Knative service after deploying the runtime image so stale pods drain.

3. **Cold-start window** — for ~200–800ms after a scaled-to-zero pod wakes up, ingress/nginx may return its own HTML before Hono is ready. The new mobile fallback handles this gracefully (stripped-tag message instead of "HTTP 404"), but the very first terminal action after sleep may show a transient error and require one retry. Not a regression — better than the current state.

4. **Ingress idle timeouts** — `bun install` can run 60s+. Confirm the staging ingress idle timeout is ≥ 120s (or the streaming response will be torn down mid-install). Worth a one-line check on the ingress config before declaring victory.

### 🚫 Out of scope (intentionally — single-scope branch)

- **Duplicate preset list** — `PRESET_COMMANDS` is now defined in both `apps/api/src/routes/terminal.ts` (local) and `packages/agent-runtime/src/runtime-terminal-routes.ts` (K8s). They will drift. Follow-up: extract to `packages/shared-runtime`.
- **Test gaps** — `/terminal/exec` happy path, `unknown_command` 400, `confirmation_required` 400, and a 401-without-token integration test should be added in a follow-up.
- **`/run` proxy buffers stdin** — uses `await c.req.text()`. Fine for terminal commands; would block on a large stdin payload. Nit, not a blocker.

---

## Risk assessment

- **Code risk:** Low. Additive routes, no schema changes, no public API changes, mirrors a router that's been in production locally for months.
- **Auth risk:** Low. Reuses the existing `deriveRuntimeToken` mechanism already in use for `/agent` and `/pool`.
- **Ops risk:** Medium until both images are deployed in the right order with stale pods drained. See checklist above.
- **Blast radius if something is wrong:** confined to the IDE terminal panel. Does not affect agent runtime, project pods' core lifecycle, or any other API surface.


ashutoshojha-odin (commented):
## Review — PR #458 (RCA + fix verification)

**TL;DR:** ✅ **Approve.** RCA is correct, fix is correct, scope is tight, tests pass locally. A few small follow-ups worth tracking but nothing blocking.

---

### RCA verification (confirmed)

I traced both the local and staging code paths and the RCA in #457 holds exactly as written:

| Path | Local | Staging (K8s) |
|---|---|---|
| `GET /api/projects/:id/terminal/commands` | `apps/api` calls in-process `terminalRoutes()` against host workspaces dir → JSON | `apps/api` proxies to `${podUrl}/terminal/commands` |
| `POST /api/projects/:id/terminal/{exec,run}` | in-process router | proxied to runtime pod |

Before this PR, the runtime pod (`packages/agent-runtime/src/server.ts`) had **no** `/terminal/*` handlers. Two failure modes resulted:

1. `GET /terminal/commands` → fell through to the SPA catch-all at `server.ts:3029` (`app.get('*', ...)`) → returned `index.html` (`text/html`). The mobile UI did `res.json()` and threw a parse error that surfaced as a generic `HTTP 404`-style banner.
2. `POST /terminal/run` and `POST /terminal/exec` → the SPA `*` handler is `GET`-only, so Hono returned a real `404`.

Both match the user-visible "HTTP 404 on staging only" symptom.

---

### Fix review (per file)

**`packages/agent-runtime/src/runtime-terminal-routes.ts`** (new, 414 lines)
- ✅ Preset list mirrors the API-side router (`bun-install`, `prisma-*`, `playwright-*`, `typecheck`, `build`) — UI contract preserved.
- ✅ `/terminal/exec` returns a `text/plain; charset=utf-8` chunked stream with the existing `[Process exited with code N]` trailer the mobile parser expects.
- ✅ `/terminal/run` emits the `\u001eSHOGO_TERM_META:<base64json>\u001e\n` sentinel — same wire format the mobile Terminal already parses for `cwd`/`exitCode`/`signal`.
- ✅ `pickCwd()` resolves relative paths against `WORKSPACE_DIR` and falls back to it when the requested cwd doesn't exist — prevents directory escape via `body.cwd`.
- ✅ `makeKillChild()` uses `process.kill(-pid, …)` against the process group (`detached: true` + group-kill) with a 2s SIGKILL escalation — handles long-running children correctly.
- ✅ Backpressure handled (`pipeWithBackpressure` pauses src on `desiredSize <= 0`).
- ✅ Dangerous-command gate (`prisma-reset`) requires `confirmDangerous: true`.

**`packages/agent-runtime/src/server.ts`**
- ✅ `authPrefixes: ['/agent', '/pool', '/terminal']` — adding `/terminal` to the auth prefix list is the right call; without it the routes would be reachable on the public preview URL anonymously.
- ✅ `app.route('/', runtimeTerminalRoutes(...))` is mounted at line **436**, well before the `app.get('*', ...)` SPA fallback at line **3029**. Mount order is correct.

**`apps/api/src/server.ts`**
- ✅ All three K8s proxy branches now forward `x-runtime-token: deriveRuntimeToken(projectId)`.
- ✅ Token symmetry verified: in `apps/api/src/lib/runtime/build-project-env.ts:76` and `knative-project-manager.ts:1051`, each pod is launched with `RUNTIME_AUTH_SECRET = deriveRuntimeToken(projectId)`. The runtime middleware in `packages/shared-runtime/src/server-framework.ts:319` does `runtimeToken === runtimeSecret`, so the forwarded header matches by string equality. ✅

**`apps/mobile/components/project/panels/ide/Terminal.tsx`**
- ✅ New `readTerminalError()` helper handles both JSON error envelopes and HTML fallthroughs (strips tags, truncates to 160 chars). This is the actual UX win — even if a future regression returns HTML again, users will see a useful message instead of a cryptic `HTTP 404`.
- ✅ Adds explicit `content-type` checks rejecting HTML on the streaming endpoints (defense in depth).
- ✅ `useState(true) → useState(false)` for `loading`: the previous initial-`true` state caused `loadCommands` to early-return on first mount because the loader had a `loading` guard. Good catch — without this fix the preset list would never load on first open.

**`packages/agent-runtime/src/__tests__/runtime-terminal-routes.test.ts`** — 2 tests covering the JSON contract on `/terminal/commands` and the streaming + meta-sentinel contract on `/terminal/run`. Ran locally:

```
bun test packages/agent-runtime/src/__tests__/runtime-terminal-routes.test.ts
✓ serves preset commands as JSON at the runtime-local path
✓ serves free-form command endpoint as a terminal stream
2 pass / 0 fail
```

---

### Non-blocking follow-ups (optional)

1. **Test coverage gaps** — would be worth adding:
   - `/terminal/exec` happy path (POST a fast preset like `typecheck` against a stub workspace) — currently only `/run` is exercised.
   - `unknown_command` 400 path.
   - `confirmation_required` 400 path for dangerous commands.
   - **Auth integration test:** assert that without `x-runtime-token` the route returns 401 in production mode. The middleware contract is the most important invariant here and the only piece that isn't covered.

2. **Duplicate preset list** — `PRESET_COMMANDS` is now defined in two places (`apps/api/src/routes/terminal.ts` for local mode and `packages/agent-runtime/src/runtime-terminal-routes.ts` for K8s). They will drift. Consider extracting to `packages/shared-runtime` or a shared package in a follow-up. Out of scope for this PR per the "single-scope branch" rule.

3. **Error envelope when HTML still leaks** — the mobile fallback strips tags and truncates to 160 chars. Fine, but on cold-start the runtime sometimes returns nginx/ingress HTML before the Hono app is up. Worth verifying that `503 / service_starting` (mentioned in the PR description) is caught earlier in the proxy and never hits this path with a confusing HTML body.

4. **`x-runtime-token` on the streaming `run` proxy** — the fetch passes `signal: c.req.raw.signal` but does **not** forward the request body as a stream (`await c.req.text()` buffers it). For typical terminal commands the body is tiny so this is fine, but if anyone ever pipes a large stdin payload through `/terminal/run` it will block. Nit, not a blocker.

---

### Rollout

PR description correctly notes both images need to ship together (API + unified runtime) — without that, the API will start sending `x-runtime-token` to a runtime that 404s, or the new runtime will reject API calls that omit the token. Coordinate the deploy.

LGTM 🚀

PR #534, opened 2026-05-12
fix(replication): auto-grant SELECT to logical_replicator on every new public table

### LINKED ISSUE #533: Remote instance stays Standby from prod — root cause: logical replication GRANTs not auto-applied to new tables
## Symptom

Desktop instances launched from a **prod-env** Studio session stay on **Standby** in the Remote Control UI and never connect. Same desktop, same network, same API key works fine from a **staging-env** session.

Desktop logs show the WebSocket failing with `code=1002 Expected 101 status code` against `wss://studio.shogo.ai/api/instances/ws` — but the WS plumbing is actually healthy (verified end-to-end against all three regional tunnel hostnames; each returns a clean `401 Authentication required` to a curl WS handshake). The actual blocker is at the **Postgres logical replication** layer: the desktop's API key row is missing from at least one region's local replica, so heartbeats routed there get `401 Invalid API key`, the desktop's `InstanceTunnel` self-heals after 3 consecutive 401s, and Remote Control flips to Standby.

## Live evidence (SigNoz, prod-india `platform-pg-1`, last 30 min)

Deduped error stream, ~one record every 10 seconds:

```
[ERROR] permission denied for table agent_eval_results
[ERROR] permission denied for table budget_alerts
[ERROR] all replication slots are in use
[ERROR] can no longer access replication slot "sub_from_eu"
[ERROR] can no longer access replication slot "sub_from_us"
[ERROR] can no longer access replication slot "sub_from_india"
[ERROR] replication slot "pg_17787_sync_<relid>_<sysid>" does not exist  (×~25 distinct slot names)
[ERROR] replication slot "pg_17788_sync_<relid>_<sysid>" does not exist  (×several more)
```

## Root cause

The publication `shogo_all_pub` is defined `FOR TABLES IN SCHEMA public`, so any new table created by `prisma migrate deploy` is auto-published. **But Postgres does NOT auto-grant `SELECT` on those new tables to the `logical_replicator` role** — and tablesync workers on each subscriber connect upstream as that role.

The cascade from one missing GRANT to a cluster-wide replication outage:

1. A migration runs as the `shogo` role: `CREATE TABLE public.agent_eval_results (...)`. The table's default ACL grants ownership/usage to `shogo` and **nothing to `logical_replicator`**.
2. Publication auto-includes the new table (via `tablesInSchema: public`).
3. `ALTER SUBSCRIPTION ... REFRESH PUBLICATION` on the next deploy registers the table on each subscriber.
4. Subscriber spawns a tablesync worker. It connects upstream as `logical_replicator` and runs `COPY public.agent_eval_results TO STDOUT`. → `permission denied for table agent_eval_results`. Worker exits.
5. apply worker respawns it. Each respawn **creates a new ephemeral replication slot** named `pg_<subid>_sync_<relid>_<sysid>`.
6. Retry storm creates slots faster than Postgres reclaims them. `max_replication_slots` exhausted → `all replication slots are in use`.
7. With the slot table full, *persistent* subscription slots (`sub_from_eu` etc.) can't be reacquired during normal reconnects → `can no longer access replication slot "sub_from_eu"`. Healthy subscriptions on the same cluster break too, freezing replication of every other table — including `api_keys`.
8. New API keys created in one region never reach the other two. Desktop's heartbeat 401s in whichever region missed replication. Remote Control = Standby.

### Why the existing RUNBOOK didn't prevent this

`k8s/cnpg/logical-replication/RUNBOOK.md` Step 6 documents the right one-time GRANT:

```sql
GRANT SELECT ON ALL TABLES IN SCHEMA public TO logical_replicator;
ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT SELECT ON TABLES TO logical_replicator;
```

…but it was run *once during initial cluster setup as `postgres`*, and **`ALTER DEFAULT PRIVILEGES` is per-grantor-role**. The unqualified form only applies to tables that `postgres` creates. Migrations don't run as `postgres` — they run as `shogo`. So `shogo`'s default privilege set is what gets consulted on a new `CREATE TABLE`, and `shogo` never had `ALTER DEFAULT PRIVILEGES ... GRANT TO logical_replicator` set. New tables → no GRANT → today's outage.

This is the same shape of bug as #502 (which manually added 26 tables to the publication before the `tablesInSchema: public` rewrite made publications self-maintaining). The publication half got the durable fix in #502 follow-ups. The GRANT half didn't, and now it's biting on the GRANT side instead.

## What needs to be done — manual remediation to clear the current outage

These three commands clear today's outage. Run in order on each region's primary. Total time: ~30 seconds per region.

### Step 1 — GRANT `logical_replicator` SELECT on every table in `public`

```bash
for ctx in oke-us oke-eu oke-india; do
  kubectl --context $ctx exec -i -n shogo-production-system platform-pg-1 -c postgres -- \
    psql -U postgres -d shogo -v ON_ERROR_STOP=1 <<'SQL'
GRANT USAGE ON SCHEMA public TO logical_replicator;
GRANT SELECT ON ALL TABLES IN SCHEMA public TO logical_replicator;
SQL
done
```

**What it does:** Adds `logical_replicator` to the ACL of every table that currently exists in `public`. Pure additive. Tables that already have this grant: no-op. Tables missing it (today: `agent_eval_results`, `budget_alerts`, and possibly others not yet erroring): the grant is now there. Tablesync workers stop hitting `permission denied`.

**Risk:** zero. Pure additive privilege change. No locks, no data touched, no restarts.

### Step 2 — Drop the leaked tablesync slots

```bash
for ctx in oke-us oke-eu oke-india; do
  kubectl --context $ctx exec -n shogo-production-system platform-pg-1 -c postgres -- \
    psql -U postgres -d shogo -v ON_ERROR_STOP=1 -c "
      SELECT slot_name, pg_drop_replication_slot(slot_name)
      FROM pg_replication_slots
      WHERE slot_name LIKE 'pg_%_sync_%' AND NOT active;
    "
done
```

**What it does:** Step 1 fixed the *cause* of the leak, but the slots that already leaked are still sitting there occupying `max_replication_slots`. This query lists every ephemeral tablesync slot that is not currently active and drops it. The persistent subscription slots (`sub_from_eu`, `sub_from_us`, `sub_from_india`) are excluded by the `pg_%_sync_%` name filter.

**Risk:** very low. `AND NOT active` filter ensures we only drop slots with no worker holding them open. The persistent subscription slots don't match the name pattern. If a tablesync was genuinely mid-flight, worst case is it restarts from scratch — no data corruption.

### Step 3 — REFRESH each subscription to retry tablesync now that GRANTs exist

```bash
kubectl --context oke-india exec -n shogo-production-system platform-pg-1 -c postgres -- \
  psql -U postgres -d shogo -v ON_ERROR_STOP=1 -c "
    ALTER SUBSCRIPTION sub_from_us  REFRESH PUBLICATION;
    ALTER SUBSCRIPTION sub_from_eu  REFRESH PUBLICATION;
  "

kubectl --context oke-eu exec -n shogo-production-system platform-pg-1 -c postgres -- \
  psql -U postgres -d shogo -v ON_ERROR_STOP=1 -c "
    ALTER SUBSCRIPTION sub_from_us    REFRESH PUBLICATION;
    ALTER SUBSCRIPTION sub_from_india REFRESH PUBLICATION;
  "

kubectl --context oke-us exec -n shogo-production-system platform-pg-1 -c postgres -- \
  psql -U postgres -d shogo -v ON_ERROR_STOP=1 -c "
    ALTER SUBSCRIPTION sub_from_eu    REFRESH PUBLICATION;
    ALTER SUBSCRIPTION sub_from_india REFRESH PUBLICATION;
  "
```

**What it does:** Forces each subscriber to re-evaluate publication membership and re-initialize the tables stuck in `srsubstate = 'i'` (initializing) or `'d'` (data-sync). With Step 1's GRANT now in place, the tablesync attempts succeed first try. Tables transition to `'r'` (ready, streaming).

**Risk:** very low. `REFRESH PUBLICATION` is the documented recovery operation for publication membership changes. Doesn't affect ongoing replication of healthy tables.

### Verification

On each region's primary:

```sql
-- 1. Slot pool back to a sane size:
SELECT slot_name, slot_type, active FROM pg_replication_slots ORDER BY slot_name;
-- Expected: only sub_from_us / sub_from_eu / sub_from_india entries, all active=t.

-- 2. Every replicated table is in 'r' state:
SELECT srsubid::regclass::text, srrelid::regclass::text, srsubstate
FROM pg_subscription_rel WHERE srsubstate <> 'r';
-- Expected: zero rows.

-- 3. logical_replicator has SELECT on every public table:
SELECT relname
FROM pg_class c JOIN pg_namespace n ON n.oid = c.relnamespace
WHERE n.nspname = 'public' AND c.relkind = 'r'
  AND NOT has_table_privilege('logical_replicator', c.oid, 'SELECT');
-- Expected: zero rows.

-- 4. api_keys count matches across regions (within seconds of lag):
SELECT count(*) FROM api_keys;
```

After Step 3 completes on the last region, the desktop's next heartbeat (within ~15 s) succeeds in whichever region Cloudflare routes it to, and Remote Control flips Standby → Online.

## What the code change does (#PR-LINK — will update below)

The manual steps clear *today*. The code change prevents the same bug from causing the next outage. Without it, the next migration to add a new table reproduces the exact cascade above — same shape, different table name.

The code change makes the GRANT a **structural property of the schema**, not a runbook step:

1. **`k8s/cnpg/logical-replication/grants.sql`** — idempotent SQL with four layers:
   - Sweep current state (`GRANT SELECT ON ALL TABLES`).
   - `ALTER DEFAULT PRIVILEGES FOR ROLE shogo` AND `FOR ROLE postgres` — covers both roles that can create tables. The next `CREATE TABLE` issued by either auto-grants `SELECT` to `logical_replicator` at create time.
   - Event trigger `auto_grant_replicator_select` on `ddl_command_end` for `CREATE TABLE` in `public`. Fires inside the database on every CREATE TABLE regardless of which role issued it. **Wrapped in `BEGIN/EXCEPTION`** so a GRANT failure raises a `WARNING` but cannot block the migration.
   - Drift audit — emits a `WARNING` listing any table that still lacks the GRANT after the sweep.

2. **`.github/workflows/deploy.yml`** — new `Reconcile logical replication GRANTs` step in each of the three regional deploy jobs (US, EU, India), placed next to the existing `Reconcile logical replication publication` step. Runs `grants.sql` via `kubectl exec` on every deploy. Same idempotent pattern. Uses `|| echo "::warning::..."` so a SQL failure cannot break a release.

### Risk profile of the code change

- **All SQL is purely additive.** No `DROP TABLE`, no `REVOKE`, no `ALTER` of existing schema objects beyond `CREATE OR REPLACE` of the trigger function.
- **The event trigger cannot block migrations.** GRANT is in `BEGIN/EXCEPTION WHEN OTHERS THEN RAISE WARNING`. If anything goes wrong, the warning lands in Postgres logs and the `CREATE TABLE` commits unaffected.
- **Idempotent.** Running 20 times = running once. Safe on every deploy.
- **Scoped strictly to `schema_name = 'public'`.** Temp tables, system catalogs, CNPG-internal schemas all unaffected.
- **No runtime impact on the api ksvc, the desktop, Knative routing, Cloudflare, or any other layer.** Only changes happen inside Postgres ACL.

## Definition of done

- [ ] Steps 1–3 run on all three regional primaries.
- [ ] Verification queries above return expected results on all three regions.
- [ ] Desktop in prod env connects and Remote Control shows Online.
- [ ] PR with `grants.sql` + `deploy.yml` change merged so the next migration doesn't reintroduce the cascade.
- [ ] (Future hardening, separate PR) CI guard in `ci-cnpg.yml` that asserts `has_table_privilege('logical_replicator', oid, 'SELECT')` is true for every `public` table in the hermetic kind cluster — fails the build on drift.

## Related

- #502 — the band-aid PR that hand-added 26 tables to the publication, surfaced the same root cause class on the publication side.
- Publication side is now durable (`tablesInSchema: public` + `bun run check:publication`). This issue is about closing the matching gap on the GRANT side.

## What I did NOT touch

The WebSocket tunnel layer is healthy and not part of this fix. Curl-tested all three regional tunnel hostnames; each returns a clean `401 Authentication required` to a real WS handshake, meaning CF passes upgrade headers, OCI LB routes correctly, Kourier matches host, `api-tunnel-direct` Service forwards to the api pod, api accepts the upgrade and rejects on auth. The earlier `503 UC` traces in SigNoz were activator cold-start moments unrelated to this bug.

ashutoshojha-odin:
Closes #533.

## TL;DR

Makes `GRANT SELECT TO logical_replicator` a structural property of the `public` schema — applied by an event trigger inside the database on every `CREATE TABLE`, with two complementary safety layers — so the next migration that adds a new table cannot reproduce the production-india replication cascade we're hitting today.

This PR does **not** clear the current outage; that needs a one-time manual remediation (kubectl + psql) documented in #533. This PR is the durable fix that stops the same cascade from recurring on the next migration.

## What was broken

The publication `shogo_all_pub` is `FOR TABLES IN SCHEMA public`, so any new table created by `prisma migrate deploy` is auto-published. But Postgres does **not** auto-grant SELECT on those new tables to the `logical_replicator` role — and tablesync workers on each subscriber connect upstream as that role. A missing GRANT causes tablesync to fail with `permission denied`, respawn endlessly, leak ephemeral replication slots, exhaust `max_replication_slots`, and then knock the *persistent* subscription slots (`sub_from_eu`, `sub_from_us`, `sub_from_india`) offline. Replication of *every* table including `api_keys` freezes. The desktop's API key row stops propagating to other regions. Heartbeats 401 in whichever region missed replication. Remote Control = Standby.

Full RCA, live SigNoz error stream, and the manual remediation runbook are in #533.

## Why the previous fixes weren't enough

- #502 fixed the same shape of bug on the *publication* side by hand-adding 26 tables (`api_keys`, `instances`, etc.).
- A follow-up rewrote the publication as `FOR TABLES IN SCHEMA public` so future tables are auto-included — durable on the publication side.
- The matching durable fix on the **GRANT** side was never written. The RUNBOOK's one-time `GRANT SELECT ON ALL TABLES` and `ALTER DEFAULT PRIVILEGES IN SCHEMA public ...` were run as `postgres`, but `ALTER DEFAULT PRIVILEGES` is **per-grantor-role** — the unqualified form only applies to tables `postgres` creates. Migrations run as `shogo`, whose default privilege set was never altered. So every new `shogo`-created table lands without the GRANT, triggering the cascade.

### `k8s/cnpg/logical-replication/grants.sql` (new, 162 lines)

Idempotent SQL with four layers, each closing a hole the others can't:

1. **Sweep** — `GRANT USAGE ON SCHEMA public` + `GRANT SELECT ON ALL TABLES IN SCHEMA public` to `logical_replicator`. Catches anything currently drifted (today: `agent_eval_results`, `budget_alerts`).
2. **Per-role default privileges** — `ALTER DEFAULT PRIVILEGES FOR ROLE shogo` AND `FOR ROLE postgres` IN SCHEMA public GRANT SELECT ON TABLES TO logical_replicator. Covers both roles that can `CREATE TABLE`. The next CREATE TABLE issued by either auto-grants at create time.
3. **Event trigger `auto_grant_replicator_select`** on `ddl_command_end` for `CREATE TABLE` / `CREATE TABLE AS` / `SELECT INTO` in `public`. Fires inside the database on every matching DDL regardless of which role issued it. The GRANT is wrapped in `BEGIN ... EXCEPTION WHEN OTHERS THEN RAISE WARNING ... END` so a failure raises a warning in PG logs but **cannot block the migration**. This is the single most important safety property of this file.
4. **Drift audit** — final `DO $$ ... $$` block emits a `WARNING` listing any table in `public` still lacking SELECT for `logical_replicator`. In a healthy cluster this never prints. Visible in deploy logs if it does.

Bootstrap-ordering safety: the first `DO $guard$` block detects the case where `logical_replicator` doesn't yet exist (very first cluster bootstrap, before CNPG's `managed.roles` reconciler has run) and `\quit`s cleanly. By the time `deploy.yml` ever invokes the script in steady state, the role exists; this branch only protects against extreme bootstrap ordering.

### `.github/workflows/deploy.yml` (+42 lines, 0 deleted)

Adds a `Reconcile logical replication GRANTs` step in each of the three regional deploy jobs (`deploy-us`, `deploy-eu`, `deploy-india`), placed immediately after the existing `Reconcile logical replication publication` step. Runs `grants.sql` via `kubectl exec -i ... -- psql -U postgres -d shogo < grants.sql`. Same idempotent pattern as the publication reconciler the team already trusts. Uses `|| echo "::warning::..."` so a SQL failure cannot break a release.

## Risk analysis

I went through every line considering "what happens if this runs in prod right now":

- **All SQL is purely additive.** No `DROP TABLE`, no `REVOKE`, no `ALTER` of existing schema objects beyond `CREATE OR REPLACE` of the trigger function (which already runs idempotently) and `DROP EVENT TRIGGER IF EXISTS` + `CREATE EVENT TRIGGER` (also idempotent).
- **The event trigger cannot block migrations.** Every GRANT call inside the function is wrapped in `BEGIN/EXCEPTION WHEN OTHERS THEN RAISE WARNING`. Bad role name, missing privilege, lock conflict — anything inside surfaces a warning to Postgres logs and returns normally. The `CREATE TABLE` that fired the trigger commits unaffected.
- **Idempotent.** Running 20 times in a row produces the same final state as running once.
- **Schema-scoped strictly to `public`.** Every filter is `schema_name = 'public'`. Temp tables (`pg_temp_*`), system catalogs, CNPG-internal schemas, and any future schemas are untouched. Matches exactly what the publication covers.
- **No runtime impact on app code, Knative routing, Cloudflare, the WS tunnel, or any layer outside Postgres.** I deliberately did not touch:
  - `apps/api/**`
  - `k8s/overlays/**` (no api-service.yaml, no api-tunnel-ingress.yaml, no DomainMapping, no Cloudflare, no Terraform)
  - The publication CR itself
  - `platform-cluster.yaml`
  - `RUNBOOK.md`, replication monitor, subscription manifests
- **deploy.yml additions are exactly three 8-line blocks** — pure additions, no removals, no edits to existing steps. `git diff --stat`: `42 insertions(+), 0 deletions(-)`.

## What runs on the next deploy after merge

For each region the new step prints:

```
NOTICE:  logical_replicator has SELECT on every table in public
GRANT
GRANT
ALTER DEFAULT PRIVILEGES
ALTER DEFAULT PRIVILEGES
CREATE FUNCTION
DROP EVENT TRIGGER
CREATE EVENT TRIGGER
```

Then the existing `Refresh logical replication subscriptions` step re-triggers tablesync for any tables previously stuck. With the GRANT now in place, tablesync succeeds first try. Slot leak stops. Slot pool drains.

## Validation done before push

- ✅ YAML parses: `deploy.yml` confirmed valid via `bun + yaml`; 10 jobs total; 3 GRANT-reconcile steps wired (one each into `deploy-us`, `deploy-eu`, `deploy-india`).
- ✅ SQL static review: every statement is idempotent, additive, and scoped to `public`. Event trigger has an `EXCEPTION` handler so it cannot block migrations.
- ⚠️ Live PG validation against a real Postgres 16/18 not run in the sandbox (no docker locally). The script is small enough that careful review is the bound. If desired, follow-up PR can wire `ci-cnpg.yml` to exercise `grants.sql` against the kind cluster's CNPG instance + a `has_table_privilege` assertion — happy to add this as a separate PR.

## Out of scope (intentionally)

- **Clearing today's outage.** That needs the manual `GRANT + drop slots + REFRESH PUBLICATION` runbook against the three live regional primaries. Full commands and verification queries are in #533. Do that first; merge this second.
- **WebSocket tunnel layer.** Curl-tested all three regional tunnel hostnames; each returns a clean `401 Authentication required` to a real WS handshake. CF + OCI LB + Kourier + `api-tunnel-direct` + api ksvc are all healthy end-to-end. The earlier `503 UC` traces in SigNoz were activator cold-start moments unrelated to this bug.
- **CI guard for GRANT drift.** Would be a great belt-and-suspenders addition to `.github/workflows/ci-cnpg.yml`: spin up the hermetic kind cluster, run the schema, then `SELECT relname FROM pg_class ... WHERE NOT has_table_privilege('logical_replicator', oid, 'SELECT')` must return zero rows. Worth doing as a follow-up, not included here to keep the surface area minimal.

## Related

- #533 — root cause analysis, manual remediation runbook
- #502 — the band-aid PR that surfaced this same root-cause class on the publication side

### DIFF
--- a/.github/workflows/ci-cnpg.yml
+++ b/.github/workflows/ci-cnpg.yml
@@ -231,6 +231,143 @@ jobs:
           done
           echo "All three regional manifests reconcile cleanly."
 
+      # ----------------------------------------------------------------------
+      # GRANT contract regression tests
+      #
+      # The publication self-maintains, but `logical_replicator` SELECT
+      # grants do NOT happen automatically. `k8s/cnpg/logical-replication/
+      # grants.sql` is supposed to close that gap via a sweep, per-role
+      # ALTER DEFAULT PRIVILEGES, and an event trigger. Three assertions:
+      #
+      #   A. After applying grants.sql, every existing table in `public`
+      #      has SELECT for `logical_replicator`.
+      #   B. A new CREATE TABLE issued AFTER grants.sql ran auto-grants
+      #      SELECT via the event trigger — no manual GRANT required.
+      #   C. **Load-bearing safety:** if the event trigger's GRANT fails
+      #      at runtime (we simulate by dropping the role), the CREATE
+      #      TABLE must still commit. The EXCEPTION handler must catch
+      #      the failure and let the migration proceed. If this ever
+      #      breaks, a future migration could be aborted by replication
+      #      machinery — strictly worse than the bug grants.sql fixes.
+      # ----------------------------------------------------------------------
+      - name: Create logical_replicator role and apply grants.sql
+        run: |
+          set -euo pipefail
+          kubectl exec -n "${NS}" platform-pg-1 -c postgres -- \
+            psql -U postgres -d shogo -v ON_ERROR_STOP=1 -c "
+              CREATE ROLE logical_replicator LOGIN REPLICATION;
+            "
+          kubectl exec -i -n "${NS}" platform-pg-1 -c postgres -- \
+            psql -U postgres -d shogo -v ON_ERROR_STOP=1 \
+              < k8s/cnpg/logical-replication/grants.sql
+
+      - name: 'Assert (A) — sweep covered every existing table'
+        run: |
+          set -euo pipefail
+          MISSING=$(kubectl exec -n "${NS}" platform-pg-1 -c postgres -- \
+            psql -U postgres -d shogo -tAc "
+              SELECT count(*)
+              FROM pg_class c
+              JOIN pg_namespace n ON n.oid = c.relnamespace
+              WHERE n.nspname = 'public' AND c.relkind = 'r'
+                AND NOT has_table_privilege('logical_replicator', c.oid, 'SELECT');
+            ")
+          if [ "${MISSING}" != "0" ]; then
+            echo "::error::sweep missed ${MISSING} tables"
+            kubectl exec -n "${NS}" platform-pg-1 -c postgres -- \
+              psql -U postgres -d shogo -c "
+                SELECT n.nspname, c.relname
+                FROM pg_class c JOIN pg_namespace n ON n.oid = c.relnamespace
+                WHERE n.nspname='public' AND c.relkind='r'
+                  AND NOT has_table_privilege('logical_replicator', c.oid, 'SELECT');
+              "
+            exit 1
+          fi
+          echo "✓ (A) all existing public tables have SELECT for logical_replicator"
+
+      - name: 'Assert (B) — event trigger auto-grants on new CREATE TABLE'
+        run: |
+          set -euo pipefail
+          # Create as `shogo` (the migration role) to ensure the trigger
+          # path is the one exercised in production, not the postgres path.
+          kubectl exec -n "${NS}" platform-pg-1 -c postgres -- \
+            psql -U postgres -d shogo -v ON_ERROR_STOP=1 -c "
+              SET ROLE shogo;
+              CREATE TABLE public.auto_grant_canary (id int PRIMARY KEY);
+              RESET ROLE;
+            "
+          OK=$(kubectl exec -n "${NS}" platform-pg-1 -c postgres -- \
+            psql -U postgres -d shogo -tAc "
+              SELECT has_table_privilege('logical_replicator', 'public.auto_grant_canary', 'SELECT');
+            ")
+          if [ "${OK}" != "t" ]; then
+            echo "::error::event trigger did NOT auto-grant SELECT on auto_grant_canary"
+            exit 1
+          fi
+          echo "✓ (B) event trigger auto-granted SELECT to logical_replicator on a fresh table"
+
+      - name: 'Assert (C) — EXCEPTION handler prevents trigger from blocking migrations'
+        run: |
+          set -euo pipefail
+          # Simulate the failure mode by deleting the role mid-flight. The
+          # next CREATE TABLE will fire the event trigger, which will try to
+          # GRANT to a role that no longer exists. Without the EXCEPTION
+          # handler this would abort the CREATE TABLE. With it, the CREATE
+          # must succeed and emit a WARNING.
+          kubectl exec -n "${NS}" platform-pg-1 -c postgres -- \
+            psql -U postgres -d shogo -v ON_ERROR_STOP=1 -c "
+              REASSIGN OWNED BY logical_replicator TO postgres;
+              DROP OWNED BY logical_replicator;
+              DROP ROLE logical_replicator;
+            "
+
+          # Run CREATE TABLE as `shogo`. Capture stderr to confirm the
+          # warning fired; exit code must be 0.
+          OUTPUT=$(kubectl exec -n "${NS}" platform-pg-1 -c postgres -- \
+            psql -U postgres -d shogo -v ON_ERROR_STOP=1 -c "
+              SET ROLE shogo;
+              CREATE TABLE public.exception_canary (id int PRIMARY KEY);
+              RESET ROLE;
+            " 2>&1)
+          EXIT_CODE=$?
+          echo "psql output:"
+          echo "${OUTPUT}"
+          if [ ${EXIT_CODE} -ne 0 ]; then
+            echo "::error::CREATE TABLE was aborted by the event trigger — EXCEPTION handler is broken."
+            echo "::error::This means a future migration could be blocked by replication machinery."
+            exit 1
+          fi
+          EXISTS=$(kubectl exec -n "${NS}" platform-pg-1 -c postgres -- \
+            psql -U postgres -d shogo -tAc \
+              "SELECT count(*) FROM pg_class c JOIN pg_namespace n ON n.oid=c.relnamespace
+               WHERE n.nspname='public' AND c.relname='exception_canary'")
+          if [ "${EXISTS}" != "1" ]; then
+            echo "::error::exception_canary table does not exist — CREATE TABLE silently rolled back"
+            exit 1
+          fi
+          # Sanity: the trigger should have emitted a WARNING about the
+          # failed GRANT. Loose match — psql formats warnings as
+          # "WARNING:  auto_grant_replicator_select: GRANT failed for ..."
+          if ! echo "${OUTPUT}" | grep -q "auto_grant_replicator_select: GRANT failed"; then
+            echo "::warning::Expected a 'GRANT failed' warning from the trigger, but did not see it in output. Trigger may not have fired."
+          fi
+          echo "✓ (C) EXCEPTION handler caught the failed GRANT; CREATE TABLE committed"
+
+      - name: 'Idempotency check — grants.sql can run twice with no error'
+        run: |
+          set -euo pipefail
+          # Restore the role so a fresh apply has a target.
+          kubectl exec -n "${NS}" platform-pg-1 -c postgres -- \
+            psql -U postgres -d shogo -v ON_ERROR_STOP=1 -c \
+              "CREATE ROLE logical_replicator LOGIN REPLICATION;"
+          for i in 1 2; do
+            echo "Run #${i}..."
+            kubectl exec -i -n "${NS}" platform-pg-1 -c postgres -- \
+              psql -U postgres -d shogo -v ON_ERROR_STOP=1 \
+                < k8s/cnpg/logical-replication/grants.sql
+          done
+          echo "✓ grants.sql is idempotent"
+
       - name: Diagnostics on failure
         if: failure()
         run: |
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1194,6 +1194,32 @@ jobs:
         run: |
           kubectl apply -f k8s/cnpg/production-us-oci/platform-publication.yaml
 
+      # ----------------------------------------------------------------------
+      # Logical replication: keep the SELECT contract for `logical_replicator`
+      # in sync.
+      #
+      # The publication is `FOR TABLES IN SCHEMA public`, so new tables join
+      # the publication automatically. But Postgres does NOT auto-grant SELECT
+      # to `logical_replicator` on those new tables — and tablesync workers
+      # connect as that role. A missing GRANT causes the tablesync worker to
+      # fail, respawn, leak a slot, and eventually break unrelated healthy
+      # subscriptions when the slot pool saturates.
+      #
+      # grants.sql installs (a) a sweep over current tables, (b) ALTER
+      # DEFAULT PRIVILEGES for every role that creates tables, and (c) an
+      # event trigger that auto-grants SELECT on every future CREATE TABLE
+      # in `public`. The event trigger wraps its GRANT in an EXCEPTION
+      # handler so a failure cannot block a migration.
+      #
+      # Idempotent and safe to run on every deploy. See the file header for
+      # the full rationale.
+      # ----------------------------------------------------------------------
+      - name: Reconcile logical replication GRANTs
+        run: |
+          kubectl exec -i -n ${{ env.NS_SYSTEM }} platform-pg-1 -c postgres -- \
+            psql -U postgres -d shogo < k8s/cnpg/logical-replication/grants.sql \
+            || echo "::warning::GRANT reconciliation failed on US (see logs above)"
+
       - name: Refresh logical replication subscriptions
         run: |
           for SUB in sub_from_eu sub_from_india; do
@@ -1420,6 +1446,14 @@ jobs:
       - name: Reconcile logical replication publication
         run: |
           kubectl apply -f k8s/cnpg/production-eu-oci/platform-publication.yaml
+
+      # GRANT reconciliation: see the matching block in the deploy-us job for
+      # full rationale. Same idempotent SQL file applied to every region.
+      - name: Reconcile logical replication GRANTs
+        run: |
+          kubectl exec -i -n ${{ env.NS_SYSTEM }} platform-pg-1 -c postgres -- \
+            psql -U postgres -d shogo < k8s/cnpg/logical-replication/grants.sql \
+            || echo "::warning::GRANT reconciliation failed on EU (see logs above)"
 
       - name: Refresh logical replication subscriptions
         run: |
@@ -1733,6 +1767,14 @@ jobs:
       - name: Reconcile logical replication publication
         run: |
           kubectl apply -f k8s/cnpg/production-india-oci/platform-publication.yaml
+
+      # GRANT reconciliation: see the matching block in the deploy-us job for
+      # full rationale. Same idempotent SQL file applied to every region.
+      - name: Reconcile logical replication GRANTs
+        run: |
+          kubectl exec -i -n ${{ env.NS_SYSTEM }} platform-pg-1 -c postgres -- \
+            psql -U postgres -d shogo < k8s/cnpg/logical-replication/grants.sql \
+            || echo "::warning::GRANT reconciliation failed on India (see logs above)"
 
       - name: Refresh logical replication subscriptions
         run: |
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1214,17 +1214,66 @@ jobs:
       # Idempotent and safe to run on every deploy. See the file header for
       # the full rationale.
       # ----------------------------------------------------------------------
+      # Resolve the *current* CNPG primary by label, not the hardcoded
+      # `platform-pg-1`. After any failover (planned switchover, OS upgrade,
+      # node drain, OOMKill) the primary may be `platform-pg-2` or
+      # `platform-pg-3`. Hardcoding `platform-pg-1` would land us on a
+      # standby where every GRANT / ALTER / CREATE fails with `read-only
+      # transaction`, the `|| echo ::warning::` would swallow it, the deploy
+      # would go green, and the contract would silently drift.
+      - name: Locate CNPG primary pod
+        id: pg_primary_us
+        run: |
+          set -euo pipefail
+          PRIMARY=$(kubectl get pods -n ${{ env.NS_SYSTEM }} \
+            -l "cnpg.io/cluster=platform-pg,cnpg.io/instanceRole=primary" \
+            -o jsonpath='{.items[0].metadata.name}')
+          if [ -z "$PRIMARY" ]; then
+            echo "::error::No primary pod found for platform-pg cluster in ${{ env.NS_SYSTEM }}"
+            exit 1
+          fi
+          echo "Located primary on US: $PRIMARY"
+          echo "PG_PRIMARY=$PRIMARY" >> "$GITHUB_ENV"
+
       - name: Reconcile logical replication GRANTs
         run: |
-          kubectl exec -i -n ${{ env.NS_SYSTEM }} platform-pg-1 -c postgres -- \
+          kubectl exec -i -n ${{ env.NS_SYSTEM }} "$PG_PRIMARY" -c postgres -- \
             psql -U postgres -d shogo < k8s/cnpg/logical-replication/grants.sql \
             || echo "::warning::GRANT reconciliation failed on US (see logs above)"
+
+      # Hard assertion: every table in `public` must have SELECT for
+      # `logical_replicator` after the reconcile step. Converts a silent
+      # warning into a failed deploy if the contract is ever drift'd.
+      - name: Verify GRANT contract
+        run: |
+          set -euo pipefail
+          MISSING=$(kubectl exec -n ${{ env.NS_SYSTEM }} "$PG_PRIMARY" -c postgres -- \
+            psql -U postgres -d shogo -tAc "
+              SELECT count(*)
+              FROM pg_class c
+              JOIN pg_namespace n ON n.oid = c.relnamespace
+              WHERE n.nspname = 'public' AND c.relkind = 'r'
+                AND NOT has_table_privilege('logical_replicator', c.oid, 'SELECT');
+            ")
+          if [ "$MISSING" != "0" ]; then
+            echo "::error::$MISSING tables in public still lack SELECT for logical_replicator on US"
+            kubectl exec -n ${{ env.NS_SYSTEM }} "$PG_PRIMARY" -c postgres -- \
+              psql -U postgres -d shogo -c "
+                SELECT n.nspname || '.' || c.relname AS missing_table
+                FROM pg_class c
+                JOIN pg_namespace n ON n.oid = c.relnamespace
+                WHERE n.nspname = 'public' AND c.relkind = 'r'
+                  AND NOT has_table_privilege('logical_replicator', c.oid, 'SELECT');
+              "
+            exit 1
+          fi
+          echo "✓ logical_replicator has SELECT on every public table on US"
 
       - name: Refresh logical replication subscriptions
         run: |
           for SUB in sub_from_eu sub_from_india; do
             echo "Refreshing $SUB on US..."
-            kubectl exec -n ${{ env.NS_SYSTEM }} platform-pg-1 -c postgres -- \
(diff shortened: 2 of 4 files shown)


ashutoshojha-odin:
## Review — overall direction approved, three substantive concerns plus nits

I agree with the framing in the PR description and #533: the publication side is durable (`FOR TABLES IN SCHEMA public` + `bun run check:publication`), and the GRANT side has been the matching hole since #502. Making GRANTs a structural property of the schema via an event trigger is the right shape of fix, not another hand-list.

That said, this PR ships with three substantive issues, one of which I'd block on and two I'd push for a follow-up. Plus nits.

---

### 🛑 Blocking — `platform-pg-1` hardcoded; will break after a CNPG failover

`deploy.yml` (all three new steps) and the *existing* `Refresh logical replication subscriptions` step that this PR sits next to both reach the database via `kubectl exec ... platform-pg-1 ...`. `platform-pg-1` is the name of the **original** primary, not the current one. After any CNPG failover (planned switchover, OS upgrade, node drain, OOMKill), the primary may be `platform-pg-2` or `platform-pg-3`. When that happens:

- The `kubectl exec` lands on a **standby**, not the primary.
- `GRANT`, `ALTER DEFAULT PRIVILEGES`, `CREATE EVENT TRIGGER` all need to run on the primary — they fail on a standby with `cannot execute ... in a read-only transaction`.
- The `|| echo "::warning::..."` swallows the failure. The deploy continues green. The GRANT contract silently drifts. **The next migration that adds a table goes back to triggering the original cascade.**

This is a latent bug already present in the publication-reconcile step; the new GRANT step inherits it. I'd want both fixed in this PR:

```bash
- name: Reconcile logical replication GRANTs
  run: |
    PRIMARY=$(kubectl get pods -n ${{ env.NS_SYSTEM }} \
      -l "cnpg.io/cluster=platform-pg,cnpg.io/instanceRole=primary" \
      -o jsonpath='{.items[0].metadata.name}')
    if [ -z "$PRIMARY" ]; then
      echo "::error::no primary found for platform-pg cluster"
      exit 1
    fi
    kubectl exec -i -n ${{ env.NS_SYSTEM }} "$PRIMARY" -c postgres -- \
      psql -U postgres -d shogo < k8s/cnpg/logical-replication/grants.sql \
      || echo "::warning::GRANT reconciliation failed on US (see logs above)"
```

Or use the read-write Service (`platform-pg-rw`) if CNPG's exec-into-Service is configured. Same fix applies to the existing publication and REFRESH steps — out of scope of this PR's *new* code but you'll want to do them in the same change because they share the bug.

The "silent failure on standby" is also why the `|| echo ::warning::` pattern by itself isn't enough — a SQL syntax error and a "wrong pod" failure both reach the same warning, and one of them means the contract is broken. Either tighten the failure mode (e.g. `exit 1` on connection/permission errors, swallow only on "already exists" type idempotency races) or add a verification step after that asserts `has_table_privilege('logical_replicator', 'public.api_keys'::regclass, 'SELECT')` returns `t`.

---

### ⚠️ Strong nit — the role-existence guard is overengineered

`grants.sql` does this dance:

```sql
DO $guard$ ...
  PERFORM set_config('grants_sql.skip', '1', false);
$guard$;
SELECT current_setting('grants_sql.skip', true) AS skip \gset
\if :{?skip}
  \if :{skip}
    \echo 'skipping: logical_replicator role missing'
    \quit
  \endif
\endif
```

Three layers of indirection (DO block → `set_config` → `current_setting → \gset` → nested `\if`) for a check that has a one-line idiom:

```sql
SELECT EXISTS (SELECT 1 FROM pg_roles WHERE rolname = 'logical_replicator') AS role_present \gset
\if :role_present
\else
  \echo 'skipping: logical_replicator role not present yet'
  \quit
\endif
```

Same behavior, half the code, no nested `\if`, no GUC-as-message-bus. The current version isn't wrong, just hard to scan. Pre-merge nit.

Also: the comment says this only fires "during an unusual bootstrap-order window," but the deploy.yml's existing `Reconcile logical replication publication` step doesn't guard against it. If this is a real concern, the publication step needs the same guard. If it's not, we can drop the guard entirely and let the script fail with a clear `role "logical_replicator" does not exist` error — the `|| echo ::warning::` already covers that. I lean toward **drop the guard**, document the contract, and let CNPG bootstrap ordering be CNPG's problem.

---

### ⚠️ Strong nit — no test for the event trigger

The `EXCEPTION WHEN OTHERS THEN RAISE WARNING` block is the load-bearing safety claim of this PR — if it doesn't actually prevent a GRANT failure from aborting a `CREATE TABLE`, this change is worse than the bug it's fixing. There's currently no automated test that verifies:

1. The trigger fires on `CREATE TABLE`, `CREATE TABLE AS`, and `SELECT INTO`.
2. The trigger grants `SELECT` to `logical_replicator` on the new table.
3. **A GRANT failure inside the trigger does not abort the outer `CREATE TABLE`** (drop `logical_replicator` mid-test, create a table, assert the create succeeded).
4. The trigger does NOT fire on temp tables (`CREATE TEMP TABLE`) or on tables in other schemas.

`ci-cnpg.yml` already spins up a hermetic kind cluster with CNPG for the publication self-maintenance test. Wiring grants.sql into the same job and asserting the four properties above is ~50 lines and would catch any regression to the EXCEPTION handler. PR description acknowledges this as a follow-up; I'd push to include it here because the EXCEPTION handler is the single safety property that lets us claim "zero risk."

---

### Nits, in order of how much they bother me

**`platform-cluster.yaml` not touched.** `bootstrap.initdb.postInitApplicationSQLRefs` is mentioned in the PR body as "future enhancement." Brand-new clusters (DR cutover, region addition) have a window between initdb and first deploy where the GRANT contract doesn't exist. If someone creates a table in that window (CNPG operator-initiated schema seeding, migration job in the bootstrap), it won't get the GRANT. Acceptable for now since the deploy step covers ongoing state, but worth a follow-up issue.

**`RUNBOOK.md` not updated.** `k8s/cnpg/logical-replication/RUNBOOK.md` Step 6 still says to run the GRANT manually. After this PR, that step is automated. Either delete Step 6 or update it to point at `grants.sql` + the deploy step. Doc drift is what got us into #502 territory — let's not start drifting again on day one.

**`ATTACH PARTITION` edge case.** The trigger fires on `CREATE TABLE PARTITION OF`, so dynamic partitions created at migrate time are covered. But `ALTER TABLE parent ATTACH PARTITION existing_table` emits `command_tag = 'ALTER TABLE'`, which the WHEN clause doesn't match. If `existing_table` was created in a non-`public` schema and then attached, it wouldn't have gotten the GRANT. We probably don't use ATTACH today; worth either a comment in grants.sql noting the gap or extending the WHEN clause to include `ALTER TABLE` (with the function reading `command_tag = 'ALTER TABLE ... ATTACH PARTITION'` specifically — slightly fiddly).

**Theoretical DROP+CREATE window.** Comment on lines 121–124 says "no concurrent DDL stream during deploy reconcile." True in steady state, but if an operator runs a `psql` against the primary concurrently with a deploy, a `CREATE TABLE` in the microseconds between `DROP EVENT TRIGGER` and `CREATE EVENT TRIGGER` would miss the GRANT. Tiny window, but the sweep step (1) catches it on the next deploy. Mention this in the comment so a future reader doesn't think it's been considered and dismissed silently.

**`ALTER DEFAULT PRIVILEGES FOR ROLE postgres` is dead code in practice.** `postgres` doesn't create tables in `public`. It's defensive, harmless, and idempotent — fine to keep, but worth a one-line comment that this is belt-and-suspenders for an unlikely operator session.

**Risk-level language in description.** "Risk: zero" is overstated. The script does modify two pre-existing object names (`grant_replicator_select_on_new_table` and `auto_grant_replicator_select`); if some unrelated extension or operator process owns objects with those names, we'd silently replace them. Astronomically unlikely, but "purely additive" is more accurate than "zero risk."

**Style nit on `grants.sql` line 65:** `SELECT current_setting('grants_sql.skip', true) AS skip \gset` — the `true` arg to `current_setting` makes it return `NULL` for an unset GUC, but the preceding DO block always sets it, so `true` is unreachable. Either drop the `true` arg or move the DO block conditional.

**Style nit on PR description:** the "Validation done before push" section says YAML lint passed and SQL static review passed; it would help reviewers to also state explicitly what was *not* tested (the EXCEPTION semantics, failover scenario, fresh-cluster initdb path). The PR body alludes to this but it's buried.

---

### What I'd want to see before approving

1. **Pod selector fixed** so the new step (and ideally the existing publication / REFRESH steps in the same job) targets the *current* primary, not `platform-pg-1` literally. Without this, the contract silently drifts after any failover and we re-create the bug we just fixed.
2. **An assertion step after the GRANT step** that fails the deploy if `has_table_privilege('logical_replicator', 'public.api_keys', 'SELECT')` is not true. This converts the warning-on-failure pattern into a guaranteed-noticed contract.
3. **At minimum a smoke test for the EXCEPTION handler** in `ci-cnpg.yml`: drop `logical_replicator`, create a table, assert the create succeeded with no error. ~20 lines, dispositive on the load-bearing safety claim.

The role-existence guard simplification, RUNBOOK update, and the other nits can be a follow-up if you want to keep this PR focused.

### What I'd green-light without changes

The four-layer architecture (sweep + per-role default privs + event trigger + drift audit) is sound. The scoping to `schema_name = 'public'` matches the publication exactly. The EXCEPTION handler is the right pattern (modulo wanting a test). The deploy.yml placement next to the existing publication reconciler is the right home.

Approving in principle, requesting changes on (1) and (2) above.

ashutoshojha-odin:
Addressed everything in `a410861b` — pushed to `fix/remote-instance`.

**Blockers**

- ✅ **Pod selector by label.** Each region's job now starts with `Locate CNPG primary pod` (resolves via `cnpg.io/cluster=platform-pg,cnpg.io/instanceRole=primary`) and exports `PG_PRIMARY` to `$GITHUB_ENV`. The new GRANT step uses `"$PG_PRIMARY"`. While I was in there I also updated the pre-existing `Refresh logical replication subscriptions` step in the same regions — same latent failover bug, fixed in the same change so the contract drift can't leak through the older path either. `grep -nE "kubectl exec.*platform-pg-1" deploy.yml` returns zero matches now (the two remaining mentions are inside the explanatory comment block).
- ✅ **Hard assertion after the reconcile.** New `Verify GRANT contract` step per region. Counts tables in `public` missing `SELECT` for `logical_replicator`, prints the offending list if non-zero, and `exit 1` — fails the deploy red instead of `|| echo ::warning::`-swallowing it.

**Strong nits**

- ✅ **EXCEPTION-handler regression test.** Five new steps appended to `cnpg-publication` in `ci-cnpg.yml`:
  - (A) sweep coverage assertion
  - (B) event trigger auto-grants SELECT on a fresh `CREATE TABLE` run as `shogo`
  - (C) **load-bearing safety**: drop `logical_replicator` mid-test (REASSIGN OWNED + DROP OWNED + DROP ROLE), then `CREATE TABLE` as `shogo` — must commit successfully. If this property ever regresses, future migrations could be aborted by replication machinery, which is strictly worse than the bug grants.sql fixes. Loose-greps the expected `auto_grant_replicator_select: GRANT failed` warning in stderr.
  - Idempotency check (grants.sql twice in a row, no error).
- ✅ **Simpler role-existence guard.** Collapsed the `DO $guard$` + `set_config`/`current_setting` + nested `\if :{?skip}` dance into:
  ```sql
  SELECT EXISTS (SELECT 1 FROM pg_roles WHERE rolname='logical_replicator') AS role_present \gset
  \if :role_present
  \else
    \echo 'skipping: logical_replicator role not present yet'
    \quit
  \endif
  ```
  Half the lines, no GUC-as-message-bus, and the unreachable `true` arg to `current_setting` is gone with the rest of the block.

**Other nits**

- ✅ `grants.sql`: added the `FOR ROLE postgres is defensive` rationale, called out the `ALTER TABLE ... ATTACH PARTITION` coverage gap explicitly, and named the theoretical microsecond DROP/CREATE window as known-mitigated-by-sweep instead of dismissed-silently.
- ✅ `RUNBOOK.md` Step 6: rewritten to describe the automation (deploy.yml + ci-cnpg.yml), with a bootstrap-only manual invocation using the new primary-selector pattern and an explicit cross-link to #533 for incident response.

**Not done, intentionally**

- `platform-cluster.yaml` `postInitApplicationSQLRefs` — the deploy.yml steps cover steady state and RUNBOOK covers the bootstrap window. Closing it structurally is worth doing but is a separate follow-up — happy to open it after this lands.
- The `ci-cnpg.yml` *diagnostics* step still references `platform-pg-1` literally. Kind cluster has one instance and cannot fail over, so this is style not correctness. Left for the postInitApplicationSQL follow-up sweep.

Net change in this push: 4 files, +334/-45. The new CI assertions will exercise on this PR — if (C) ever goes red, do NOT merge: it means the EXCEPTION handler has regressed and a migration could be aborted by the trigger.

PR #577, opened 2026-05-17
fix(worker,agent-runtime,studio): support cli-worker as a self-hosted…

sunilyanagandula:
## Description (Required)

When a Studio user picks a self-hosted Shogo worker (a `cli-worker` instance) from the machine picker, the UI currently 502s on every request to `/api/projects` and the agent-runtime binary itself crashes with `ENOENT: tree-sitter.wasm` because the compiled binary has the build-machine's filesystem path baked into it. This PR fixes both issues, in two commits, so a self-hosted operator gets a working Studio + agent flow end-to-end.

### Commit 1 — Support cli-worker as a self-hosted execution target

Studio's `RemoteAwareHttpClient` was written assuming any remote instance is a data source (true for desktop, false for cli-worker). Every `/api/projects/*` request was routed through cloud's transparent proxy → worker tunnel → 502, because the cli-worker's tunnel only serves `/agent/*` paths. Three coordinated patches:

- **Patch A — Workspace seeding.** `WorkerRuntimeManager` now always sets `projectDir` and fails loudly with a multi-line operator-facing diagnostic when neither `autoPull` nor a pre-pulled `projectsDir` provides a real workspace. The worker tunnel returns a structured 502 (`{code, message, path}`) instead of a generic 502 when `resolveLocalUrl` rejects a request.
- **Patch B — Bundle `runtime-template/` next to the binary.** Post-compile script copies the Vite/React scaffolding into `dist/`, and `getRuntimeTemplatePath()` looks at `dirname(process.execPath)/runtime-template` second (after the `RUNTIME_TEMPLATE_DIR` env override).
- **Patch C — Studio data-API gating.** `useActiveInstance` and `useRemoteState` now consult `instance.kind`. Only `desktop` instances get a `remoteProxyBaseUrl`; `cli_worker` instances leave it null so Studio falls through to the local Cloud DB. `remoteAgentBaseUrl` is unchanged for both kinds, so agent execution still tunnels through the cli-worker correctly.

Drive-by build fix: added `--external electron` to all `build:bin:*` scripts. Empirically derived as the only flag actually needed at Bun 1.3.13 (the other three flags from `progress.md` are not required, dropped per minimal-set rule).

### Commit 2 — Bundle tree-sitter WASMs alongside compiled binary

`bun build --compile` resolves `__dirname + '/tree-sitter.wasm'` at build time and bakes the build-machine filesystem path into the binary. A Mac-built agent-runtime references `/Users/<dev>/.../tree-sitter.wasm` and ENOENT-aborts when an operator runs the same artifact on a Linux VPS.

- **Post-compile bundling.** `scripts/copy-tree-sitter-wasm-to-dist.ts` copies `web-tree-sitter`'s parser core and 7 language grammars (matching `code-extractor.ts:EXTENSION_TO_LANGUAGE`) into `dist/tree-sitter-wasm/`.
- **Runtime resolution.** `code-extractor.ts:getWasmDir()` picks the first reachable directory: `$TREE_SITTER_WASM_DIR` → `dirname(process.execPath)/tree-sitter-wasm` → source-tree fallback. Always wires `Parser.init`'s `locateFile` so `web-tree-sitter` never falls back to its baked Mac path.
- **Belt-and-suspenders worker env.** `WorkerRuntimeManager.buildEnv` exports `TREE_SITTER_WASM_DIR=dirname(runtimeBinPath)/tree-sitter-wasm` to the spawned runtime by default. Operator overrides on the host AND per-project `extraEnv` still win in that order.

**Distribution implication.** Prebuilt agent-runtime is now binary + sibling `runtime-template/` + sibling `tree-sitter-wasm/` together. `shogo runtime install` (separate fix) must ship all three.

### Out of scope

While testing I observed a pre-existing race in `WorkerRuntimeManager.maybeAutoPull` (on `main`, not introduced here): `pulledProjects.add(projectId)` is called before the git clone completes, so a second concurrent inbound request can spawn a runtime against a half-cloned workspace. The runtime then seeds its template, which collides with git's still-in-flight checkout. The current fallback chain (template seeding) keeps the worker functional even when this race fires, so it's noisy rather than blocking. Filed as a follow-up — straightforward fix is gating `pulledProjects.add` behind clone completion or wrapping the body in a per-projectId mutex.

## Related Issue (Required)

commit 1: https://odin-ai.atlassian.net/browse/EN-3772
commit 2: https://odin-ai.atlassian.net/browse/EN-3773

## Type of Change (Required)

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update
- [ ] Refactoring (no functional changes, no test changes)
- [x] Added new tests

### Automated

- `packages/shogo-worker/src/lib/__tests__/runtime-manager-auto-pull.test.ts` — auto-pull happy paths + loud failure when misconfigured
- `packages/shogo-worker/src/lib/__tests__/runtime-manager-describe-rejection.test.ts` — structured 502 codes for every rejection path
- `packages/shogo-worker/src/lib/__tests__/tunnel-structured-502.test.ts` — tunnel sends `{code, message, path}` JSON
- `packages/shogo-worker/src/lib/__tests__/runtime-manager-tree-sitter-env.test.ts` — `TREE_SITTER_WASM_DIR` injection + override precedence
- `packages/agent-runtime/src/__tests__/binary-ships-runtime-template.integration.test.ts` — Patch B post-compile layout + execPath resolution
- `packages/agent-runtime/src/__tests__/binary-ships-tree-sitter-wasms.integration.test.ts` — WASM post-compile layout, byte-identical-to-source guarantee, dead-weight prevention
- `packages/agent-runtime/src/__tests__/tree-sitter-wasm-resolution.test.ts` — resolver priority order
- `packages/shared-app/src/hooks/__tests__/computeRemoteProxyBaseUrl.test.ts` — Patch C gating logic

All added/modified tests pass; no pre-existing tests regressed.

### Manual — Stage 1 (Mac, darwin-arm64)

1. `bun run build:bin:darwin-arm64` succeeds, produces `dist/{shogo-agent-runtime-darwin-arm64, runtime-template/, tree-sitter-wasm/}`.
2. Copied artifacts to `~/.shogo/runtime/`, started `bun run dev:all` (local Studio on `:8081`, API on `:8002`, SQLite).
3. Started worker: `bun packages/shogo-worker/src/cli.ts worker start --cloud-url http://localhost:8002 --project <uuid> --foreground`.
4. **Patch A verified**: auto-pull tried git → fell back to file-transport → fell back to template seeding. No crash, runtime spawned with valid `WORKSPACE_DIR`.
5. **Patch B verified**: log shows `[workspace-defaults] Seeded runtime template into workspace` from the bundled location.
6. **Patch C verified**: in browser DevTools, `GET http://localhost:8002/api/projects` returned 200 directly — no `/instances/<id>/p/` proxy URL. Same for `/api/projects/<id>` and the sidebar's other data-API calls.
7. **Patch C inverse verified**: `/api/projects/<id>/agent-proxy/agent/chat/<sessionId>/turn` correctly went through the cloud proxy (uses `remoteAgentBaseUrl`, which is preserved for cli-worker by design).
8. Sent a chat message; agent ran a full LLM turn (`[AgentChat] Background stream completed`). No tree-sitter ENOENT errors.

### Manual — Stage 2 (Linux x86-64 VPS)

1. `bun run build:bin:linux-x64` produced an `ELF 64-bit LSB executable, x86-64` binary plus the two sidecar dirs.
2. SCP'd binary + `runtime-template/` + `tree-sitter-wasm/` to `/root/.shogo/runtime/` on the VPS. Compiled `shogo-linux-x64` worker CLI to `/usr/local/bin/shogo`.
3. Exposed local Cloud through ngrok, added the ngrok URL to `ALLOWED_ORIGINS` in `.env.local`.
4. Started worker on VPS: `SHOGO_API_KEY=... ANTHROPIC_API_KEY=... shogo worker start --cloud-url <ngrok-url> --project <uuid> --foreground`.
5. **Blocker 4 verified on the actual Linux target**: binary booted with no `tree-sitter.wasm` ENOENT and no `runtime-template/*` ENOENT. Tunnel connected through ngrok.
6. **End-to-end agent verified on Linux**: sent a chat message in Studio, agent runtime accepted the request, LLM turn completed, response rendered in browser. Status bar showed `Running on srv1643414` confirming traffic routed to the VPS instance.
7. Browser DevTools confirmed the same Patch C gating behaviour as Stage 1: data API direct, agent path proxied.

### How to reproduce locally

1. Check out branch: `git fetch origin && git checkout fix/self-hosted-worker-cli-worker-support`
2. `bun install && bun run dev:all` (boots local Cloud on SQLite).
3. In `packages/agent-runtime/`: `bun run build:bin:darwin-arm64`.
4. Copy `dist/{shogo-agent-runtime-darwin-arm64 → agent-runtime, runtime-template, tree-sitter-wasm}` to `~/.shogo/runtime/`. `chmod +x ~/.shogo/runtime/agent-runtime`.
5. Create a project in local Studio at `http://localhost:8081`, copy the project UUID and an API key.
6. From repo root: `SHOGO_API_KEY=<key> bun packages/shogo-worker/src/cli.ts worker start --cloud-url http://localhost:8002 --project <uuid> --foreground`.
7. Switch instance picker to your local machine, open the project. Sidebar loads with no 502s. Send a chat — agent responds (requires `ANTHROPIC_API_KEY` in `.env.local`).


sunilyanagandula:
I have read the CLA Document and I hereby sign the CLA

PR #612, opened 2026-05-20
fix(canvas): auto-reload iframe on rebuild instead of showing a manual "Refresh" toast

### LINKED ISSUE #611: Canvas iframe doesn't auto-refresh on rebuild — shows a manual "Refresh" toast instead of reloading
## Summary

The **Canvas preview does not update live** when chat changes files, and on a brand-new project the canvas stays empty even after the agent creates the first surface. The only ways the content actually swaps in are three workarounds: switching tabs inside the project (which unmounts/remounts the iframe), refreshing the whole page, or noticing and clicking the small "Refresh" toast that appears inside the iframe.

Linked Jira: [SHOG-654](https://odin-ai.atlassian.net/browse/SHOG-654).

## Reproduction

**Case A — first build (most common report)**

1. Create a brand-new project.
2. Ask chat to "build a dashboard with a bar chart of revenue."
3. Wait for the agent to finish writing files and report success.

**Expected:** the dashboard appears in the canvas within a second of the agent finishing.
**Actual:** the canvas stays empty. A small "Update available — Refresh" toast may be visible at the bottom of the iframe (often clipped on smaller layouts so users miss it). Switching to a different project tab and back, or refreshing the page, finally renders the dashboard.

**Case B — subsequent edits**

1. Inside a project with an existing canvas surface, ask chat to "change the chart color to amber."
2. Watch the canvas while the agent edits files.

**Expected:** the chart re-renders amber within a beat of the agent's last write.
**Actual:** the chart stays the old color. Same three workarounds required to see the change.

## Root cause

The live-refresh pipeline is **fully wired end-to-end**. The break is exactly one design choice in the bridge:

| Step | Location | Status |
|---|---|---|
| File watcher broadcasts `{ type: 'reload' }` when files change | `packages/agent-runtime/src/canvas-file-watcher.ts:236` | ✅ |
| `GET /agent/canvas/stream` streams the event over SSE; replays an `init` event on connect | `packages/agent-runtime/src/server.ts:3404` | ✅ |
| `canvas-bridge.js` receives the event inside the iframe → calls `showUpdateToast()` | `packages/agent-runtime/static/canvas-bridge.js:162` (old) | ❌ |
| `showUpdateToast()` injects a `<div>` with a manual "Refresh" button — it does **not** reload the document on its own | `canvas-bridge.js:138–167` (old) | ❌ |

The new Vite bundle is on disk and the iframe knows about it. The user-visible React tree, however, is still bound to the **old** (or, on a first build, **empty**) bundle. Until something forces a real navigation, nothing renders.

### Why each workaround appears to "fix" it

- **Switching project tabs.** `<CanvasWebView />` unmounts the iframe when its tab is hidden. Coming back remounts it → a fresh HTTP fetch picks up the new bundle.
- **Page refresh.** Same end result via the browser-level navigation.
- **Clicking the toast's Refresh button.** Calls `window.location.reload()`. Works — but most users never see the toast because it renders **inside the iframe**, often below the visible canvas viewport.

### Why first-build is the worst case

On a brand-new project, the canvas iframe loads against an empty workspace. React mounts against **zero surfaces** — there is literally nothing to render. The agent then writes the first surface and fires `reload`. With the old handler, the iframe stays bound to the empty initial bundle, so the user reads "canvas didn't build the first time" when in reality the build succeeded but the swap never happened.

## Fix (root cause, not a patch)

Replace the manual-toast handler with a properly engineered auto-reload pipeline in the same file. See PR for the implementation.

Guards in the fix:

1. **Debounce ~250ms.** A single rebuild fan-outs into multiple file-watcher events. We collapse them into one reload after the burst settles.
2. **Keep the `ready` gate** from the server's `init` replay so the very first connect can't reload-loop.
3. **Visibility gate.** If `document.hidden`, defer the reload until `visibilitychange` fires. Backgrounded canvases stop thrashing.
4. **`reloadInFlight` flag.** Events arriving during the navigation are dropped — reloads never stack.
5. **Transient "Updating…" pill** (no manual buttons) for visual affordance so the swap isn't a mystery flash.
6. **`requestAnimationFrame` before `location.reload()`** so the pill actually paints before the document tears down.

## Why this layer (not `CanvasWebView.tsx`, not the SSE event)

- `canvas-bridge.js` is **served verbatim by the runtime** — no build step. Every existing canvas picks up the change on its next iframe load. No template re-seed, no per-project rebuild, no migration window.
- Bumping `refreshKey` from `<CanvasWebView />` to force a remount would split responsibility across two files, lose the debounce/visibility logic, and contradict the existing comment contract (`// Reload is handled inside the iframe itself — main.tsx listens to the SSE stream on the same origin.`).
- Renaming the SSE event (e.g. `autoreload`) would be a server change for purely cosmetic reasons. The existing `reload` already means "the bundle is new, reload now" — only the bridge was misinterpreting it.

## Acceptance criteria

- [ ] Chat creating a new surface on a fresh project renders that surface in the canvas within ~300ms, with no manual interaction.
- [ ] Chat editing an existing surface shows the change within ~300ms.
- [ ] A burst of consecutive file writes results in exactly **one** reload, ~250ms after the last write.
- [ ] If the canvas tab is hidden, the reload defers and runs the moment it becomes visible again.
- [ ] No reload-loop on initial mount (the server's `init` replay must not be interpreted as a `reload`).
- [ ] The existing tab-switch and page-refresh workarounds still work (no regression).

## Risk

- Single file, plain JS, `+64 / -25`. No schema, no API contract change, no migration.
- Worst-case regression: auto-reload throws synchronously → behavior degrades to today's bug. Strictly non-regressing.
- Canvas is an agent-driven preview, not an IDE — full reload is the expected mental model. HMR-style state preservation is a separate design conversation, not in scope.

## Out of scope (follow-ups)

- HMR-style soft updates that preserve React state across rebuilds (requires a Vite HMR client inside the canvas, not just SSE).
- Per-surface granular reload (current contract is whole-document reload; future optimization).

ashutoshojha-odin:
Closes #611. Linked Jira: [SHOG-654](https://odin-ai.atlassian.net/browse/SHOG-654).

## TL;DR

One file changed (`packages/agent-runtime/static/canvas-bridge.js`, `+64 / -25`, plain JS, no build step). Replaces a manual "Refresh" toast with a properly engineered auto-reload pipeline so the canvas updates live when chat changes files — instead of requiring the user to switch tabs, refresh the page, or click an easy-to-miss toast inside the iframe.

## The bug

The live-refresh pipeline was already complete from agent → file watcher → SSE → iframe:

| Step | Location | Status |
|---|---|---|
| File watcher broadcasts `{ type: 'reload' }` on file changes | `packages/agent-runtime/src/canvas-file-watcher.ts:236` | ✅ |
| `GET /agent/canvas/stream` streams the event; replays `init` on connect | `packages/agent-runtime/src/server.ts:3404` | ✅ |
| `canvas-bridge.js` receives the event → `showUpdateToast()` | `canvas-bridge.js:162` (old) | ❌ |
| `showUpdateToast()` injects a `<div>` with a manual "Refresh" button — **does not reload** | `canvas-bridge.js:138–167` (old) | ❌ |

The new Vite bundle was on disk, the event was delivered, the iframe knew. But `window.location.reload()` only fired if the user clicked a button rendered inside the iframe — frequently clipped on smaller layouts.

**First-build symptom (the worst case):** iframe loads against an empty workspace, React mounts against zero surfaces, agent then writes the first surface and fires `reload`, bridge shows the toast, user sees "canvas is empty" and assumes the build failed.

**Edit symptom:** chat says "updated the dashboard," agent did succeed, but the canvas keeps showing the old UI until the user notices the toast, switches project tabs (which unmounts/remounts the iframe), or refreshes the page.

## The fix

Replace the toast-and-wait handler with an auto-reload pipeline in the same file:

```js
// SSE listener — rebuild events from the agent runtime
var RELOAD_DEBOUNCE_MS = 250
var reloadTimer = null
var reloadPending = false
var reloadInFlight = false

function scheduleReload() {
  if (reloadInFlight) return
  reloadPending = true
  if (reloadTimer) clearTimeout(reloadTimer)
  reloadTimer = setTimeout(performReloadIfVisible, RELOAD_DEBOUNCE_MS)
}

function performReloadIfVisible() {
  if (!reloadPending) return
  if (document.hidden) return // visibilitychange will retry
  reloadPending = false
  reloadInFlight = true
  showUpdatingPill()
  requestAnimationFrame(function () { window.location.reload() })
}

document.addEventListener('visibilitychange', function () {
  if (!document.hidden && reloadPending) performReloadIfVisible()
})

var es = new EventSource('/agent/canvas/stream')
var ready = false
es.onmessage = function (e) {
  var evt = JSON.parse(e.data)
  if (evt && evt.type === 'init') { ready = true; return }
  if (evt && evt.type === 'reload' && ready) scheduleReload()
}
```

### The six guards (and why each one exists)

1. **Debounce 250ms.** A single rebuild fan-outs into multiple file-watcher events (write the file → fsync → emit). We collapse them into one reload after the burst settles.
2. **`ready` gate.** The server replays an `init` event on every connect. Without this gate the very first message would be interpreted as `reload` and we'd loop on mount.
3. **Visibility gate.** If `document.hidden` (canvas tab in the project layout is hidden, OS window backgrounded, etc.) defer the reload until `visibilitychange`. Stops backgrounded canvases from thrashing on every agent file write.
4. **`reloadInFlight` flag.** Events arriving during the navigation are dropped — reloads never stack. Without this, a long-running rebuild that fires multiple `reload`s during the navigation could queue extra reloads on top of the next document.
5. **Transient "Updating…" pill** (no manual buttons). Visual affordance so the swap isn't a mystery flash. We removed the old "Refresh" / dismiss buttons entirely — they were workarounds for the bug, not features.
6. **`requestAnimationFrame` before `location.reload()`.** The pill needs to actually paint before the document tears down. Without rAF, navigation can fire in the same frame and the pill never renders.

## Why this layer (and not the alternatives)

I considered three places to fix this and chose the bridge for concrete, defensible reasons:

1. **`canvas-bridge.js` (this PR).** Plain JS served verbatim by the runtime — no build step. **Every existing project picks up the fix on the next iframe load.** No template re-seed, no per-project rebuild, no migration window. ✅
2. **Bump `refreshKey` from `<CanvasWebView />` to force a remount.** Rejected. Splits responsibility across two files, loses the debounce/visibility logic, and contradicts the existing contract comment in `CanvasWebView.tsx`: `// Reload is handled inside the iframe itself — main.tsx listens to the SSE stream on the same origin.`
3. **Rename the SSE event** (e.g. `reload` → `autoreload`). Rejected. Server change for purely cosmetic reasons. The existing `reload` already means "the bundle is new, reload now" — only the bridge was misinterpreting it.

## Verification

- ✅ `bun --print "new Function(require('fs').readFileSync('packages/agent-runtime/static/canvas-bridge.js','utf8')); 'JS OK'"` parses clean.
- ✅ Existing migration test `packages/agent-runtime/src/__tests__/canvas-bridge-migration.test.ts` passes.
- ✅ `grep` confirms no lingering references to the removed `showUpdateToast`/`refresh-btn`/`dismiss-btn` strings anywhere in the repo.
- ✅ Manual reproduction of all three observable behaviors (first-build, edit, debounced fan-out) demonstrably fixed.

## Test plan for reviewer

1. **First-build case.** On `main`: create a new project, ask chat to build a dashboard → canvas stays empty until you switch tabs. Switch to this branch → canvas renders within ~300ms of the agent's last write.
2. **Edit case.** Open an existing project, ask chat to change a color. On `main` you'll need to click the toast or switch tabs. On this branch the change paints automatically.
3. **Debounce.** Trigger a rebuild that writes 5–10 files in quick succession. Open DevTools → Network → confirm exactly **one** document reload, ~250ms after the last file write.
4. **Visibility gate.** Background the OS window (or click into another project tab) while a rebuild is happening. No reload. Bring the canvas tab back to focus → reload fires immediately.
5. **No reload-loop on mount.** Hard-refresh the canvas. The `init` event arrives → no spurious reload follows.
6. **Backwards compat.** Tab-switch and page-refresh still work as fallback paths.

## Risk

- **Surface:** single file, 89 lines touched. No schema, no API contract change, no migration.
- **Failure mode:** if `scheduleReload`/`performReloadIfVisible` throw synchronously, behavior degrades to **today's** bug — the user gets no auto-reload but the workarounds still work. **Strictly non-regressing.**
- **State preservation:** the canvas is an agent-driven preview, not an IDE. Full reload is the expected mental model. If we later need HMR-style state preservation that's a separate design conversation (would require a Vite HMR client inside the canvas, not just SSE).
- **Blast radius beyond this PR:** none. The handler is local to the iframe; no consumer-facing API changes.

## Out of scope (intentional)

- HMR-style soft updates preserving React state across rebuilds.
- Per-surface granular reload (current contract is whole-document; future optimization).
- UI to disable live-refresh per project (would only be needed if a user case for it surfaces — current behavior matches user expectation universally).

## Checklist

- [x] Linked to GitHub issue (#611) and Jira (SHOG-654)
- [x] One-file change, plain JS, no build step
- [x] All six guards implemented and documented inline
- [x] Bridge migration test passes
- [x] No lingering references to removed `showUpdateToast` / button strings
- [x] Manual repro of first-build, edit, and debounce cases all green
- [x] Failure path is non-regressing
- [x] Comment block in code explicitly marks the handler as the single source of truth for live-refresh so future contributors don't add a parallel remount path

### DIFF
--- a/packages/agent-runtime/static/canvas-bridge.js
+++ b/packages/agent-runtime/static/canvas-bridge.js
@@ -132,41 +132,80 @@
     document.head.appendChild(style)
   }
 
-  function showUpdateToast() {
-    // Cross-script dedup: if any prior bridge (e.g. an old embedded copy
-    // baked into a stale main.tsx) already mounted a toast, don't add a
-    // second one.
+  // Transient "Updating…" pill shown *while* the auto-reload is in flight.
+  // We deliberately no longer expose a manual "Refresh" button — the canvas
+  // is agent-driven, so the moment a rebuild lands the user expects to see
+  // it. The pill exists purely for affordance so the swap isn't a
+  // mysterious flash.
+  function showUpdatingPill() {
     if (document.getElementById(TOAST_ID)) return
     injectToastStyles()
-
     var toastEl = document.createElement('div')
     toastEl.id = TOAST_ID
-
     var label = document.createElement('span')
-    label.textContent = 'Update available'
-
-    var refreshBtn = document.createElement('button')
-    refreshBtn.className = 'refresh-btn'
-    refreshBtn.textContent = 'Refresh'
-    refreshBtn.onclick = function () { window.location.reload() }
-
-    var dismissBtn = document.createElement('button')
-    dismissBtn.className = 'dismiss-btn'
-    dismissBtn.textContent = '\u00d7'
-    dismissBtn.onclick = function () {
-      var el = document.getElementById(TOAST_ID)
-      if (el) el.remove()
-    }
-
+    label.textContent = 'Updating\u2026'
     toastEl.appendChild(label)
-    toastEl.appendChild(refreshBtn)
-    toastEl.appendChild(dismissBtn)
     document.body.appendChild(toastEl)
   }
 
   // -------------------------------------------------------------------------
   // SSE listener — rebuild events from the agent runtime
   // -------------------------------------------------------------------------
+  //
+  // Behavior:
+  //   1. Server replays an `init` event on connect — gate live updates on it
+  //      so the very first message can't double-reload.
+  //   2. On `reload`, debounce ~250ms (a single rebuild can fan out into
+  //      multiple file-watcher events) then `window.location.reload()`.
+  //   3. If the tab is hidden (e.g. user is on a different IDE tab inside
+  //      the canvas), defer the reload until visibility returns so we don't
+  //      thrash backgrounded previews.
+  //   4. Show a transient "Updating…" pill while the reload is in flight so
+  //      the swap has an affordance and doesn't feel like a random flash.
+  //
+  // This is the source of truth for live-refresh — the parent <CanvasWebView />
+  // explicitly does NOT remount the iframe on rebuild (see comment in
+  // CanvasWebView.tsx). All other refresh paths (tab switch unmounting the
+  // iframe, manual page refresh) were workarounds for this handler showing a
+  // manual "Refresh" toast instead of actually reloading. They still work,
+  // but should no longer be necessary.
+
+  var RELOAD_DEBOUNCE_MS = 250
+  var reloadTimer = null
+  var reloadPending = false
+  var reloadInFlight = false
+
+  function scheduleReload() {
+    if (reloadInFlight) return
+    reloadPending = true
+    if (reloadTimer) clearTimeout(reloadTimer)
+    reloadTimer = setTimeout(performReloadIfVisible, RELOAD_DEBOUNCE_MS)
+  }
+
+  function performReloadIfVisible() {
+    if (!reloadPending) return
+    if (typeof document !== 'undefined' && document.hidden) {
+      // Wait for the tab to come back into focus — visibilitychange handler
+      // below will call us again.
+      return
+    }
+    reloadPending = false
+    reloadInFlight = true
+    showUpdatingPill()
+    // Defer one frame so the pill paints before the navigation tears down
+    // the document.
+    if (typeof requestAnimationFrame === 'function') {
+      requestAnimationFrame(function () { window.location.reload() })
+    } else {
+      window.location.reload()
+    }
+  }
+
+  if (typeof document !== 'undefined' && document.addEventListener) {
+    document.addEventListener('visibilitychange', function () {
+      if (!document.hidden && reloadPending) performReloadIfVisible()
+    })
+  }
 
   try {
     var es = new EventSource('/agent/canvas/stream')
@@ -175,11 +214,11 @@
       try {
         var evt = JSON.parse(e.data)
         if (evt && evt.type === 'init') { ready = true; return }
-        if (evt && evt.type === 'reload' && ready) showUpdateToast()
+        if (evt && evt.type === 'reload' && ready) scheduleReload()
       } catch (_err) { /* ignore malformed events */ }
     }
   } catch (_err) {
-    // Older browsers without EventSource: degrade gracefully (no live toasts).
+    // Older browsers without EventSource: degrade gracefully (no live reload).
   }
 
   // -------------------------------------------------------------------------
--- a/packages/agent-runtime/static/canvas-bridge.js
+++ b/packages/agent-runtime/static/canvas-bridge.js
@@ -20,7 +20,24 @@
 // ---------------------------------------------------------------------------
 
 (function () {
-  if (window.__shogoCanvasBridgeLoaded) return
+  if (window.__shogoCanvasBridgeLoaded) {
+    // A second copy of the bridge is trying to load — usually because a stale
+    // main.tsx is shipping its own embedded copy alongside the live-served
+    // /agent/canvas/bridge.js, or because the parent <CanvasWebView /> grew a
+    // separate remount-on-rebuild path. Either way, two reload handlers on the
+    // same document will fight each other (double reloads, racing state).
+    // Surface it loudly in dev so it's caught at review time instead of in
+    // production telemetry six months later.
+    if (typeof console !== 'undefined' && console.warn) {
+      console.warn(
+        '[shogo-canvas-bridge] Duplicate bridge load detected. ' +
+        'Live-refresh is the sole responsibility of /agent/canvas/bridge.js — ' +
+        'remove the duplicate (likely an embedded copy in main.tsx or a parent-side ' +
+        'remount on rebuild). See CanvasWebView.tsx for the contract.'
+      )
+    }
+    return
+  }
   window.__shogoCanvasBridgeLoaded = true
 
   // -------------------------------------------------------------------------
@@ -136,14 +153,19 @@
   // We deliberately no longer expose a manual "Refresh" button — the canvas
   // is agent-driven, so the moment a rebuild lands the user expects to see
   // it. The pill exists purely for affordance so the swap isn't a
-  // mysterious flash.
-  function showUpdatingPill() {
+  // mysterious flash. `bufferedCount` lets us surface "N changes while you
+  // were away" when emerging from a hidden state.
+  function showUpdatingPill(bufferedCount) {
     if (document.getElementById(TOAST_ID)) return
     injectToastStyles()
     var toastEl = document.createElement('div')
     toastEl.id = TOAST_ID
     var label = document.createElement('span')
-    label.textContent = 'Updating\u2026'
+    if (bufferedCount && bufferedCount > 1) {
+      label.textContent = 'Updating\u2026 (' + bufferedCount + ' changes)'
+    } else {
+      label.textContent = 'Updating\u2026'
+    }
     toastEl.appendChild(label)
     document.body.appendChild(toastEl)
   }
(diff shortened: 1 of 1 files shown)


ashutoshojha-odin:
## Follow-up commit: `b51a9552` — defer-on-activity, ready-gate fallback, duplicate-bridge warning

Layered six engineering guards on top of the original auto-reload fix, directly addressing the realistic drawbacks I'd identified:

| # | Concern | Mitigation in `b51a9552` |
|---|---|---|
| 2 | Reload yanks the page out from under in-flight clicks/typing | Capture-phase passive listeners on `mousedown`/`keydown`/`touchstart`/`pointerdown`/`wheel` track `lastUserInputAt`. `performReloadIfReady()` defers an additional `USER_IDLE_MS` (1500ms) if the user is currently interacting, then re-tries. |
| 4 | `ready` gate could be silently muted if the server contract ever changes | `READY_FALLBACK_MS` (5000ms) — if `init` never arrives, flip `ready = true` anyway with a `console.warn` for triage. Bridge can never get permanently muted. |
| 7 | No context for the user when emerging from a hidden state with multiple buffered reloads | `bufferedReloadCount` is incremented on every `scheduleReload()`, and the pill now reads `Updating… (N changes)` when N > 1. |
| 9 | Contract enforced only by a code comment — future contributor could add a parallel remount path in `CanvasWebView.tsx` | Duplicate-load detection in the IIFE guard now `console.warn`s loudly explaining the contract and where to look. |
| 10 | No telemetry / triage signal for "canvas didn't update" reports | Read-only debug handle at `window.__shogoCanvasBridge` exposing `ready`, `reloadPending`, `reloadInFlight`, `bufferedReloadCount`, `lastUserInputAt`, `userActive`, `pausedByUser`, `documentHidden`. Users can paste it from the iframe console. |
| 11 | No opt-out for demos / presentations | `localStorage.setItem('shogo:canvas:pauseReload', '1')` pauses live-refresh entirely. Cleared by removing the key. Honored at the top of `scheduleReload()` — events are dropped, not queued. |

Still the same file (`canvas-bridge.js`), still plain JS, no build step. Existing `canvas-bridge-migration.test.ts` still passes.

## Known limitations — filed as follow-ups

Three concerns were real but out of scope for a bridge-layer fix. Filed as Jira tickets so they aren't forgotten:

- **[SHOG-656](https://odin-ai.atlassian.net/browse/SHOG-656)** — Preserve React state across rebuilds via a Vite HMR client. The current full-reload model is fine for previews but will surface as data loss the moment a canvas surface has a form or multi-step wizard.
- **[SHOG-657](https://odin-ai.atlassian.net/browse/SHOG-657)** — Parent-driven visibility signal for React Native WebView. `document.hidden` is unreliable across iOS/Android `react-native-webview` versions; `CanvasWebView` should `postMessage` visibility transitions so the gate works on native.
- **[SHOG-658](https://odin-ai.atlassian.net/browse/SHOG-658)** — Subscribe `canvas-file-watcher` to Vite's `bundleEnd` hook instead of raw fs events. Today the 250ms debounce covers up the race because incremental builds are usually <250ms, but it's coincidence not contract.

## What stays as-is (intentional)

- **DevTools state loss on reload** — unfixable at this layer; this is what reload means.
- **Reload during an in-flight `fetch()`** — covered in practice by the new defer-on-activity gate (any user-initiated POST involves a click/keypress within the previous 1500ms).
- **`reloadPending` lifetime while backgrounded for hours** — desired behavior; the new pill counter at least makes it visible.

## Reviewer test plan (additions)

On top of the original test plan:

7. **Defer on activity.** While typing in any canvas input, have chat write a file. Confirm the reload waits ~1.5s after your last keystroke before firing. `window.__shogoCanvasBridge.userActive` should be `true` during the deferral.
8. **Pause for demo.** `localStorage.setItem('shogo:canvas:pauseReload', '1')` in the iframe console. Trigger rebuilds. Confirm no auto-reload happens. Remove the key, trigger another rebuild — auto-reload resumes.
9. **Buffered-count pill.** Background the canvas tab, trigger 3+ rebuilds, bring it back. Pill should read `Updating… (N changes)` for one frame before the navigation.
10. **Debug handle.** Run `window.__shogoCanvasBridge` in the iframe console. Returns a plain object with the listed shape. Modifications to the returned object don't affect bridge state (it's a getter).
11. **Duplicate-bridge warning.** Manually inject a second `<script>` pointing at `/agent/canvas/bridge.js` after the first has loaded. Confirm a single `console.warn` with the contract message; behavior is otherwise unchanged.

## Diff summary

```
 packages/agent-runtime/static/canvas-bridge.js | 153 +++++++++++++++++++++----
 1 file changed, 129 insertions(+), 24 deletions(-)
```

Nothing else touched.

PR #619, opened 2026-05-20
fix(agent-runtime): touch runtime on AI-proxy auth (defence-in-depth keepalive)

halfd0rk:
## Why

PR #591 keeps the local agent runtime warm while bytes are flowing on `/agent/*` proxied streams. That covers the common SSE-chatty case but misses one corner: a long Opus / GPT-4o agent turn whose only outbound traffic for several minutes is one large model call goes silent on `/agent/*` for the duration of that call — and on a single-user dev machine, silent for >15min trips `RUNTIME_IDLE_MS` and SIGTERMs the runtime mid-completion.

That's the failure mode that surfaces in the console as:

```
[api] [AgentProxy] GET /agent/canvas/stream failed after 24 attempts: Unable to connect.
```

right after a `[WorkerRuntimeManager] idle-evicting … after 900s` line.

## What

Touch the locally-managed runtime on every successful AI-proxy auth.

The strongest possible signal that an agent runtime is alive and busy is that the agent inside it just successfully authenticated an `/ai/v1/*` or `/ai/anthropic/v1/*` request — i.e. it is actively making a model call right now. Touching there closes the gap left by #591 without any per-chunk overhead on the model-proxy hot path.

Implementation:

- New `touchRuntimeFor(payload)` helper inside `aiProxyRoutes()`:
  - No-op for `payload === null` (auth failed).
  - No-op for workspace-scoped Shogo API keys (`payload.projectId === 'api-key'`) — there's no local runtime to touch.
  - `try`/`catch` around `getRuntimeManager().touch(...)` so a missing/mid-construction `RuntimeManager` (cloud pod / k8s mode) never fails an auth.
- `validateProxyAuth` and `validateAnthropicAuth` are split into thin wrappers around `*Impl` bodies. The wrappers call `touchRuntimeFor(payload)` after a successful resolve. Auth semantics are unchanged — a `null` payload still maps to `null` and the route returns 401.

## Stacking

This depends on `RuntimeManager.touch()` / `IRuntimeManager.touch()` from #591. Base branch is set to `fix/agent-runtime-keepalive-while-streaming`; once #591 lands, GitHub will retarget this to `main`.

## Test plan

- [ ] Existing `runtime-manager-touch.test.ts` still passes.
- [ ] `bun test apps/api/src/__tests__/` clean.
- [ ] Local repro: start a long Anthropic / OpenAI completion (>5 min, e.g. via `/ai/v1/chat/completions` with `stream: true`) and confirm `[WorkerRuntimeManager] idle-evicting …` no longer fires during the call. The stream completes without the AgentProxy reconnect storm.
- [ ] Cloud-pod mode (no local runtime): the `try`/`catch` around `getRuntimeManager().touch()` swallows; auth still works.
- [ ] Workspace API key path (`shogo_sk_*`): `touchRuntimeFor` short-circuits on `'api-key'`; no surprise `touch` calls.

### DIFF
--- a/apps/api/src/routes/ai-proxy.ts
+++ b/apps/api/src/routes/ai-proxy.ts
@@ -38,6 +38,7 @@ import { verifyRuntimeToken } from '../lib/runtime-token'
 import { resolveApiKey } from './api-keys'
 import { wipeCloudKey } from '../lib/cloud-key-wipe'
 import { getShogoCloudUrl } from '../lib/cloud-urls'
+import { getRuntimeManager } from '../lib/runtime'
 import {
   MODEL_CATALOG,
   MODEL_ALIASES,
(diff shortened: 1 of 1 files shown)


halfd0rk:
Superseded by 5f4cea65 on main, which already ships the AI-proxy touch hook (and the heartbeat from #591, and the idleMs<=0 / local-mode guard). Closing without merge — main is already correct.

PR #830, opened 2026-07-22
Feat/tooltips homepage

meghakawad-debug:
## Summary
Updated the web/desktop tooltip experience and refreshed tech stack descriptions so users get clearer guidance when choosing modes, actions, environments, models, and project tech stacks.
## Changes
- Added/shared web tooltip support across chat and project controls.
- Added tooltip labels for key compact/full chat input actions, including:
  - Interaction mode
  - Stakeholder summary toggle
  - Quick actions
  - Model picker
  - Attach files
  - Voice project creation
  - Environment picker
  - Project source/menu actions
- Fixed desktop/web tooltip behavior so only one tooltip appears on hover.
  - Removed the native browser `title` tooltip from the custom tooltip wrapper.
  - Preserved accessibility with `aria-label`.
- Updated tech stack descriptions across runtime stack metadata to make stack choices clearer.
- Updated desktop dev startup to sync the web bundle before launching Electron, so desktop fallback loads the latest tooltip changes.
## Testing
- [x] Tested locally
- [ ] Added/updated tests
- [x] Verified no regressions
Tested with:
- Verified tooltip source no longer uses native `title={label}`.
- Verified custom tooltip uses `aria-label`.
- Synced desktop embedded web bundle with `npm --prefix apps/desktop run sync:web`.
- Ran `git diff --check`.
- Ran `npm --prefix apps/desktop run build`.

Screenshot 2026-07-22 at 2 24 39 PM
Screenshot 2026-07-22 at 2 24 42 PM

### DIFF
--- a/apps/mobile/components/chat/CompactChatInput.tsx
+++ b/apps/mobile/components/chat/CompactChatInput.tsx
@@ -64,7 +64,7 @@ import {
 } from "../../hooks/useTypingPlaceholder"
 
 import { AttachSourceSheet } from "./AttachSourceSheet"
-import { WebTooltip } from "../ui/web-tooltip"
+import { WebTooltip } from "../ui/tooltip"
 
 const MAX_FILE_SIZE = 10 * 1024 * 1024
 const MAX_FILES = 10
--- a/apps/mobile/components/chat/CompactChatInput.tsx
+++ b/apps/mobile/components/chat/CompactChatInput.tsx
@@ -64,28 +64,13 @@ import {
 } from "../../hooks/useTypingPlaceholder"
 
 import { AttachSourceSheet } from "./AttachSourceSheet"
+import { WebTooltip } from "../ui/web-tooltip"
 
 const MAX_FILE_SIZE = 10 * 1024 * 1024
 const MAX_FILES = 10
 
 const MIN_INPUT_HEIGHT = 80
 const MAX_INPUT_HEIGHT = 200
-
-/**
- * Show a native browser tooltip on hover (web only). Wraps children in a
- * `display: contents` div with the `title` attribute so layout is unaffected
- * and the trigger's own ref (e.g. for popover positioning) isn't disturbed.
- * On native this is a transparent passthrough — the icon click opens the
- * popover menu which already shows the full label.
- */
-function WebTooltip({ label, children }: { label: string; children: React.ReactNode }) {
-  if (Platform.OS !== "web") return <>{children}</>
-  return React.createElement(
-    "div",
-    { title: label, style: { display: "contents" } },
-    children,
-  )
-}
 
 interface AttachedFile {
   id: string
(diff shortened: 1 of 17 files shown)


Adhi15 (approved):

apps/mobile/components/chat/CompactChatInput.tsx
```diff
@@ -64,29 +64,14 @@ import {
 } from "../../hooks/useTypingPlaceholder"
 
 import { AttachSourceSheet } from "./AttachSourceSheet"
+import { WebTooltip } from "../ui/tooltip"
```
Adhi15, line 68: Please verify that the imported `WebTooltip` from `../ui/tooltip` actually accepts a `placement` prop — if it's not wired up in that component, these additions silently do nothing.

apps/mobile/components/chat/CompactChatInput.tsx
```diff
@@ -790,16 +775,18 @@ export const CompactChatInput = forwardRef<View, CompactChatInputProps>(
...
-                    {...triggerProps}
-                    disabled={disabled}
-                    className="h-[22px] flex-row items-center gap-1 rounded-md px-1.5"
-                  >
-                    <Text className="text-xs text-muted-foreground">
-                      {resolveShortName(currentModelId)}
-                    </Text>
-                    <ChevronDown className="h-2 w-2 text-muted-foreground/60" size={8} />
-                  </Pressable>
+                  <WebTooltip label={`Model: ${resolveShortName(currentModelId)}`} placement="bottom">
```
Adhi15, line 778: The `Pressable` here has `{...triggerProps}` from the popover — please verify on web that wrapping it in `WebTooltip` doesn't intercept pointer events or break the popover trigger (the old local `WebTooltip` used `display: contents` specifically to avoid this).

apps/mobile/components/chat/CompactChatInput.tsx
meghakawad-debug: Resolved. I verified that the shared `WebTooltip` from `../ui/tooltip` accepts a `placement` prop and now passes it directly through to the underlying shared tooltip primitive:
```tsx
<Tooltip placement={placement} ...>

apps/mobile/components/chat/CompactChatInput.tsx
meghakawad-debug: Resolved. I updated `WebTooltip` so it no longer adds an extra hit-test wrapper around the `Pressable`.
For valid React element children, the tooltip now clones the child and merges the tooltip trigger props/ref/events onto the actual child element. That keeps the popover `Pressable` as the clickable target, so the existing `{...triggerProps}` behavior is preserved.
There is only a `display: contents` fallback for non-element children, matching the previous intent of avoiding pointer-event/layout interference.

Adhi15 (approved):

PR #882, opened 2026-09-15
ci(ios): deploy main builds to production TestFlight

ashutoshojha-odin:
## Summary

Implements [SHOG-745](https://odin-ai.atlassian.net/browse/SHOG-745).

The task is to make every push to `main` produce a full iOS build and upload it to TestFlight, while ensuring that the app is built against the production backend and keeps the same marketing version as the production release. Main-branch builds must not silently use staging or advance to a new app version.

## Previous behavior

- The iOS workflow could skip a `main` push when the changed files did not match the mobile path filter.
- A `main` push resolved to the `staging` environment.
- JavaScript-only changes could take the EAS OTA update path when the native fingerprint was already cached.
- The staging path bundled `https://studio.staging.shogo.ai`.
- Main builds could fall back to the placeholder version in `apps/mobile/app.json` instead of using the production release version.

## What changed

- Main-branch pushes now bypass the mobile path gate, so every push continues into the iOS pipeline.
- Main now resolves to the `production` environment.
- Main pushes always set `needs_build=true`, which forces the full native Xcode build instead of the OTA-only path.
- Production API URL resolution now applies to main builds, so the JavaScript bundle receives `https://studio.shogo.ai`.
- Main builds resolve their marketing version from the latest existing semver `v*` production tag. The current repository production tag is `v1.14.6`, so main builds use marketing version `1.14.6`.
- The resolver fails closed when no production `v*` tag exists instead of silently using the placeholder `apps/mobile/app.json` version.

## How the workflow works now

1. A push to `main` triggers `.github/workflows/ios.yml`.
2. The `gate` job detects that the event is a `push` to `main` and writes `skipped=false` to its job output.
3. The fingerprint job resolves `environment=production`.
4. The version resolver checks the existing Git tags and selects the current semver `v*` production tag. It removes the `v` prefix and exposes the result as the app version output.
5. The build decision checks the main-branch condition before the fingerprint cache condition and forces `needs_build=true`.
6. The OTA update job is skipped for this event.
7. The native build job updates the temporary CI copy of `apps/mobile/app.json` with the production marketing version and assigns a new iOS build number from the GitHub Actions run number.
8. Expo prebuild and Xcode bundling receive production environment variables, including `EXPO_PUBLIC_API_URL=https://studio.shogo.ai` and `EXPO_PUBLIC_APP_ENV=production_app`.
9. Xcode archives and signs the iOS app using the configured Apple certificate and provisioning profile.
10. Because the resolved environment is production, the submit decision is always enabled. The IPA is uploaded to App Store Connect/TestFlight using the existing Apple credentials.

The marketing version remains stable across normal main pushes. Only the iOS build number changes per workflow run, allowing multiple TestFlight builds for the same production release version. A new marketing version is used only after a new production `v*` tag exists.

## Files changed

- `.github/workflows/ios.yml`
  - Main pushes always pass the gate.
  - Main resolves to production.
  - Main pushes always force a full native build and TestFlight submission.
- `scripts/ci/resolve-mobile-app-version.sh`
  - Main/production builds use the existing production `v*` tag.
  - Production version resolution refuses to fall back to `apps/mobile/app.json` when no tag exists.

## What remains unchanged

- Direct `v*` and `ios-v*` tag pushes continue to trigger production iOS builds.
- Staging remains available through manual workflow dispatch.
- Android workflows are unchanged.
- Backend deployment workflows are unchanged.
- The existing signing, archive, IPA artifact, and App Store Connect upload steps are reused.

## Test plan

- [x] `git diff --check`
- [x] Bash syntax checks for the CI scripts
- [x] YAML parse for `.github/workflows/ios.yml`
- [x] Main-branch version resolution returns `1.14.6` from `v1.14.6`
- [x] Production tag version resolution returns `1.14.6`
- [x] Working tree is clean after the commit and push
- [x] Draft PR targets `main`
- [ ] GitHub macOS archive and TestFlight upload smoke test

## Validation

Local workflow and resolver checks passed. The final archive and TestFlight upload require the GitHub Actions macOS runner, Apple signing secrets, and App Store Connect credentials.


lacvapps (approved):
Reviewed and approved for merge (Lint, Typecheck, Test all passing; CLA check bypassed by admin decision).

ashutoshojha-odin:
I have read the CLA Document and I hereby sign the CLA

PR #903, opened 2026-09-18
feat(tier2): Docker-compose project class — Phases 0–3 (metal VM classes, exposed ports, tunnel, public preview)

lacvapps:
## What

Tier 2 (hosted Docker-capable project class) plan, Phases 0–3.

**New `docker-compose` tech stack** (`packages/agent-runtime/tech-stacks/docker-compose`)
- `stack.json` declares `runtime.vmClass: 'docker'`, `runtime.minimumInstanceSize: 'large'`, `runtime.installCommand`, `runtime.ports`.
- Working starter: FastAPI app + Postgres via `docker-compose.yml`, plus `.shogo/STACK.md` agent guidance.
- `TechStackMeta` (agent-runtime) and the tech-stack registry (`packages/core`) both extended with `vmClass`/`minimumInstanceSize`/`ports`, mirrored + drift-checked (apps/api can't read `stack.json` directly, same reason `target` is mirrored today).

**PreviewManager**: new `compose` devServer kind. This is intentionally *not* lifecycle-managed — no readiness polling against `docker compose ps`, no log streaming, no quiesce/resume hooks. It just stops the stack from silently falling through to the `vite` default (a pre-existing bug for any unrecognized `devServer` value) and lets the agent run `docker compose` itself via `exec`.

**permission-engine**: `docker` / `docker compose` / `make` join the Balanced-mode allowlist, but only when `SHOGO_RUNTIME_CLASS=docker`. `sudo` stays hard-blocked in every mode regardless — dockerd runs as root from `fc-init`, the agent's user only needs group membership.

**API plumbing**
- `runtime-class-setting.ts`: platform gate `runtime.docker_class_enabled`, mirroring the existing `sandbox-exec-setting.ts` pattern 1:1, + `GET`/`PUT /api/admin/settings/docker-class` (protected by the existing `/api/admin/settings/*` super-admin middleware).
- `instance-sizes.ts`: `applyDockerStackFloor()` — unlike the mobile floor, this one is **billed** (dockerd + a real compose stack need genuine CPU/RAM).
- `build-project-env.ts`: derives `SHOGO_RUNTIME_CLASS=docker` from the project's tech stack (`isDockerTechStack`), re-checking the platform gate on every assignment; falls back to a standard VM with a loud `console.error` if the gate is off. Also injects `SHOGO_EXPOSED_PORTS` (comma-separated declared ports) for the guest-side port allowlist.

The `runtime.docker_class_enabled` gate defaults **off**, so this changes nothing for any existing project unless a project explicitly opts into the `docker-compose` stack.

## Phase 0 — feasibility spike (done, on real staging bare metal)

Ran the Phase 0 feasibility spike directly on the live staging metal host (`latitude-dal-1`), against the exact guest kernel already deployed to the fleet (`vmlinux-6.1.102`). Fully isolated from the live warm pool (standalone `firecracker` processes, a tap device outside the `fctap<n>` naming the pool's GC matches on, throwaway rootfs copies); host verified back to its exact pre-spike state afterward. Full writeup: [`docs/runbooks/docker-project-class-phase0-spike.md`](https://github.com/shogo-labs/shogo-ai/blob/feat/tier2-docker-project-class/docs/runbooks/docker-project-class-phase0-spike.md).

**Headline result: dockerd works on the stock fleet kernel with zero kernel changes.** overlay2, cgroup v2, bridge/veth, and IP forwarding are all already there. One fix was needed: dockerd's default `iptables-nft` backend fails in-guest; `update-alternatives --set iptables /usr/sbin/iptables-legacy` resolves it with no kernel dependency.

Validated with the actual PR starter (FastAPI + Postgres, real `docker compose up --build`, real healthchecks) end to end, including **suspend/resume with live containers**, using the real production mechanism — `Pause` → full snapshot-to-disk → kill the process → recreate the tap → fresh `firecracker` process → `LoadSnapshot resume_vm=true`. The app kept serving real DB-backed requests immediately after restore into a brand-new process.

## Phase 1 — metal-agent internals + class-aware placement

- **VM classes** (`apps/metal-agent/src/config.ts`): `standard` / `docker`, each with its own base rootfs override, min instance size, and support gating (`isVmClassSupported`).
- **Second data drive** (`data-drive.ts`, new): a dedicated `virtio-blk` device for `/var/lib/docker`, sized independently of the CoW root, provisioned/released alongside the VM's lifecycle.
- **Class-aware warm pools** (`pool.ts`, `firecracker-vm-manager.ts`, `snapshot-store.ts`, `cache-index.ts`, `rootfs.ts`, `fc-api.ts`): `metal-agent` now maintains separate pools per class, with class-scoped rootfs identity (`computeRootfsIdentity` accepts a `baseRootfsOverride`) so a docker-class snapshot is never served to a standard-class request or vice versa. `reclaimOrphans` releases data-drive files alongside CoW/rootfs cleanup.
- **Heartbeat + placement**: `register.ts` reports per-class pool capacity (`classes: [{ vmClass, supported, poolSize, available }]`) in the agent's heartbeat; `routes/metal.ts` ingests it; `metal-warm-pool-controller.ts` prefers a docker-capable host for a docker-class assignment while preserving stickiness for an already-placed project and falling back safely to a standard-only host if none is available.

## Phase 3 — exposed ports, client-side tunnel, public per-port preview

New inbound network surface, built defaults-off and opt-in per port:

- **Data model** (`lib/project-ports.ts`, new): a project's tech stack declares a fixed, non-extensible set of ports (e.g. `docker-compose` declares `8000` http/preview, `5432` tcp/tunnel). `Project.settings.exposedPorts` can only toggle an already-declared port's `visibility` (`'tunnel'` | `'preview'`), never add one.
- **REST surface**: `GET /api/projects/:id/ports` (list, with `previewUrl` for any port currently `preview`-visible) and `PATCH /api/projects/:id/ports/:port` (toggle visibility; rejects `preview` on a non-`http` port).
- **Guest-side bridging** (`packages/agent-runtime/src/port-bridge.ts`, new): `/agent/ports/:port/http/*` (HTTP reverse proxy to `127.0.0.1:{port}`) and `/agent/ports/:port/ws` (raw TCP-over-WebSocket bridge), both gated by the `SHOGO_EXPOSED_PORTS` allowlist as defense-in-depth even though the caller is already authenticated.
- **Client-side TCP tunnel** (`apps/api/src/lib/port-tunnel-bridge.ts`, new + WS route on `/api/projects/:id/ports/:port/tunnel`): desktop/CLI ↔ API ↔ project runtime, for any declared port regardless of protocol/visibility. Authenticated via `shogo_sk_*` API key or session cookie (the WS upgrade never reaches Hono middleware, so this is re-implemented directly against the same credential shapes `authMiddleware` accepts).
- **Public per-port preview**: `{port}--{projectId}.preview.<base>` hostname → new API render proxy `/api/preview/:id/ports/:port/render` (anonymous, like the existing root preview proxy, but additionally gated on the port being declared, `protocol: 'http'`, and currently `visibility: 'preview'`) → forwards to the runtime's `/agent/ports/:port/http/*` with a server-derived runtime token. The Cloudflare Worker (`terraform/modules/preview-router/main.tf`) parses the new hostname shape and routes it straight to the port-render endpoint — no Knative anchor fallback, since docker-class projects only ever run on metal.

## Testing

Full suites (isolated per-file where the package has an isolated runner, to avoid `mock.module()` bleed across files):

- `apps/metal-agent`: 429 pass, 0 fail.
- `packages/agent-runtime`: 3934 pass, 0 fail relevant to this change (20 pre-existing failures, all external-API-dependent: expired/invalid Composio & Serper keys, an OpenAI param change, one env-leak flake — none touch changed files).
- `packages/core`: 138 pass, 0 fail.
- `packages/shared-runtime`: 675 pass, 0 fail.
- `apps/api`: 9161 pass, 84 fail — 77 are the same pre-existing external-API/mock-bleed flakes noted above; the remaining 7 are top-level duplicate test files that require a live Postgres connection in this environment (their `lib/__tests__/` counterparts, which are mocked, pass cleanly) — pre-existing and unrelated to this change.

New unit tests: `project-ports.test.ts`, `port-bridge.test.ts`, `port-tunnel-bridge.test.ts`, `config.test.ts` / `data-drive.test.ts` / `pool-vm-class.test.ts` (metal-agent), plus new cases in `metal-warm-pool-controller.test.ts`, `build-project-env.test.ts`, and `preview-worker-routing.integration.test.ts` (per-port Worker routing).

Next: security review of the new inbound surface (tunnel WS auth, port-render anonymity boundary, guest-side allowlist) before any rollout.

Made with [Cursor](https://cursor.com)


lacvapps:
## Security review

Ran a security review of this diff focused on the new inbound network surface: the port-tunnel WS auth (`authenticatePortTunnelWs`), the anonymous per-port preview render proxy, the guest-side port-bridge allowlist, the exposed-ports data model's project scoping, the docker-class admin gate, and metal VM-class placement/isolation.

**No medium-or-higher findings** — the changed code enforces the intended boundaries (declared-port allowlists, protocol/visibility gates, workspace/session auth, super-admin-only gate, fail-closed VM class fallback).

Closed out the four low-severity defense-in-depth notes it raised (`64345812`):
- `settings.runtimeClass === 'docker'` now requires the tech stack to actually agree before granting the docker-class VM (no exploitable path today — no write site for that field — but tightened before one exists).
- `previewPortRenderHandler` now strips `x-runtime-token`/`authorization` from the runtime's response before it reaches the anonymous public preview visitor.
- Per-port preview routes now get the same `previewFrameStrip` treatment (drop `X-Frame-Options`/COOP) as the root preview proxy, so they can be embedded the same way.
- `PORT_TUNNEL_WS_PATH_RE`'s port capture is now as strict as the guest-side `parsePortParam()` (no leading zeros).

Also fixed an unrelated `tsc --noEmit` type error surfaced while re-checking (`getDeclaredPorts()`'s return type was inferred as possibly-`undefined`) — pure type-level fix, no behavior change.

